PR #29, opened 2020-05-24
No boolean literal compare rule

dkrutskikh:


lib/src/rules/no_boolean_literal_compare_rule.dart
```diff
@@ -0,0 +1,85 @@
...
+      bool useDirect;
+
+      if (expression.leftOperand is! BooleanLiteral) {
+        correction = expression.leftOperand.toString();
+        useDirect = (expression.operator.type == TokenType.EQ_EQ &&
+                expression.rightOperand.toString() == 'true') ||
+            (expression.operator.type == TokenType.BANG_EQ &&
+                expression.rightOperand.toString() == 'false');
+      } else {
+        correction = expression.rightOperand.toString();
```
Anrock, line 46: Identical to then branch, except for right/left operand. Maybe extract this logic to subfunction and just pass expression inside?

test/rules/no_boolean_literal_compare_rule_test.dart
```diff
@@ -0,0 +1,111 @@
...
+
+  if (b != d) {}
+
+  [true, false].where((value) => value == true).where((value) => value == c);
+}
+
+''';
+
+void main() {
+  test('NoBooleanLiteralCompareRule report about founded issues', () {
```
Anrock, line 42: `report about founded` => `reports about found`

PR #30, opened 2020-06-03
Newline before return

dkrutskikh:


lib/src/rules/no_boolean_literal_compare_rule.dart
```diff
@@ -10,6 +10,12 @@ import 'rule_utils.dart';
 // Inspired by TSLint (https://palantir.github.io/tslint/rules/no-boolean-literal-compare/)
 
 class NoBooleanLiteralCompareRule extends BaseRule {
+  static const _failureComprareNullAwarePropertyWithTrue =
+      "Compare boolean null aware property with boolean literal 'true' is not semantic recommended.";
```
Anrock, line 14: `Comparison of null-conditional boolean with boolean literal may result in comparing null with boolean.`

lib/src/rules/no_boolean_literal_compare_rule.dart
```diff
@@ -10,6 +10,12 @@ import 'rule_utils.dart';
...
+
+  static const _correctionComprareNullAwarePropertyWithTrue =
+      "Replace operator equality and 'true' literal with operator if null and 'false' literal";
```
Anrock, line 17: `Prefer using null-coalescing operator with false literal on right hand side`.

PR #31, opened 2020-06-05
No empty block

dkrutskikh:


tool/format_coverage.dart
```diff
@@ -8,22 +8,23 @@ void main() {
   final lcovReportFile = File('coverage/coverage.lcov');
   if (!lcovReportFile.existsSync()) {
     _printCoverageOutputDoesntExistBanner();
-    return;
-  }
+  } else {
```
Anrock, line 11: Why increase indent level here? IMO, it was better with early return before.

doc/rules/no_empty_block.md
```diff
@@ -0,0 +1,38 @@
+# No empty block
+
+## Rule id
+no-empty-block
+
+## Description
+Disallows empty blocks. Blocks with a todo comment inside are not considered empty.
```
Anrock, line 7: How about treating blocks with any comment as non-empty? Majority of empty blocks are empty catch blocks which are intentionally empty so there is no sense in adding todo comment there but "intentionally do nothing"-comments seem to fit fine. Or there should be an option to not report empty catch blocks, just like in TSLint.

lib/src/rules/no_empty_block.dart
```diff
@@ -0,0 +1,49 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:dart_code_metrics/src/models/code_issue.dart';
+import 'package:dart_code_metrics/src/models/code_issue_severity.dart';
+
+import 'base_rule.dart';
+import 'rule_utils.dart';
+
+// Inspired by TSLint (https://palantir.github.io/tslint/rules/no-empty/)
```
Anrock, line 9: Should go to docs, imo.

lib/src/analysis_options.dart
```diff
@@ -1,6 +1,8 @@
 import 'package:meta/meta.dart';
 import 'package:yaml/yaml.dart';
 
+// Documantation about customizing static analysis located at https://dart.dev/guides/language/analysis-options
```
Anrock, line 4: This should go into README, otherwise not really discoverable.

PR #33, opened 2020-06-07
add avoid-preserve-whitespace-false rule

andrewst:
@dkrutskikh  add add avoid-preserve-whitespace-false rule

### DIFF
--- a/lib/src/rules/avoid_preserve_whitespace_false.dart
+++ b/lib/src/rules/avoid_preserve_whitespace_false.dart
@@ -1,0 +1,56 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:dart_code_metrics/src/models/code_issue.dart';
+import 'package:dart_code_metrics/src/models/code_issue_severity.dart';
+
+import 'base_rule.dart';
+import 'rule_utils.dart';
+
+class AvoidPreserveWhitespaceFalseRule extends BaseRule {
+  static const _failure =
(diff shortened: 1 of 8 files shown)


lib/src/rules/avoid_preserve_whitespace_false.dart
```diff
@@ -0,0 +1,56 @@
...
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:dart_code_metrics/src/models/code_issue.dart';
+import 'package:dart_code_metrics/src/models/code_issue_severity.dart';
+
+import 'base_rule.dart';
+import 'rule_utils.dart';
+
+class AvoidPreserveWhitespaceFalseRule extends BaseRule {
+  static const _failure =
+      'Avoid use preserveWhitespace: false.';
```
Anrock: `Avoid using`.

doc/rules/avoid_preserve_whitespace_false.md
```diff
@@ -0,0 +1,40 @@
+# Double literal format
+
+## Rule id
+avoid_preserve_whitespace_false
+
+## Description
+Avoid use argument `preserveWhitespace: false` for `@Component` annotation in Angular,
```
Anrock, line 7: `Avoid setting preserveWhitespace in Angular @Component annotations to false explicitly. Its default value is already false.`?

lib/src/rules/avoid_preserve_whitespace_false.dart
andrewst: fixed

doc/rules/avoid_preserve_whitespace_false.md
andrewst: fixed

PR #52, opened 2020-06-23
Binary expression operand order

Anrock:


dkrutskikh:
@Anrock please register new rule in lib/src/rules_factory.dart

PR #71, opened 2020-07-18
Members ordering rule implementation

### LINKED ISSUE #37: Implement member-ordering rule
Allows to enforce declaration order in classes. For example:

`member-ordering: [angular-inputs, angular-outputs, public, constructor, private]`

Inspired by [TSLint rule](https://palantir.github.io/tslint/rules/member-ordering/)

incendial:
@dkrutskikh can u please take a look?
Also, are there any rules for change log update?

closes #37

### DIFF
--- a/doc/rules/member_ordering.md
+++ b/doc/rules/member_ordering.md
@@ -11,14 +11,14 @@ Enforces member ordering.
 The value for order may be an array consisting of the following strings:
 
 - public_fields
-- private_field
+- private_fields
 - public_getters
 - private_getters
 - public_setters
 - private_setters
 - public_methods
 - private_methods
-- constructor
+- constructors
 - angular_inputs
 - angular_outputs
 - angular_host_bindings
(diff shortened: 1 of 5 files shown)


Anrock:
@incendial Order should be configurable via .analysis_options by user, but I've failed to find any code related to that. I think without configuration this rule would be quite useless for end-users.

incendial:
@Anrock 
> I think without configuration this rule would be quite useless for end-users
It seems that it as useless as other (because none of them is configurable, especially `no-magic-number`) so for now we can decide default order and update the rule when generic configuration mechanism will be added. What do u think?
Also @dkrutskikh said that there are some ideas for rules configuration, are they on your roadmap?

doc/rules/member_ordering.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Rule id
+
+member-ordering
+
+## Description
+
+Enforces member ordering.
+
+The value for order may be an array consisting of the following strings:
```
Anrock, line 11: Don't forget to remove this, if we gonna merge with configuration.

incendial:
@Anrock @dkrutskikh I have added config support for the rule, please review new version

PR #72, opened 2020-07-19
Prefer conditional expressions rule implementation

### LINKED ISSUE #41: Implement prefer-conditional-expressions rule
Issue a warning when there is a if-else block and same variable is written to in both branches.

Bad:
```dart
var myVar;
if (someTest) {
  myVar = a;
} else {
  myVar = b;
}
```

Good:
```dart
var myVar = someTest ? a : b;
```

Inspired by [TSLint rule](https://palantir.github.io/tslint/rules/prefer-conditional-expression/)

incendial:
@dkrutskikh can u please take a look?

Closes #41

### DIFF
--- a/lib/src/rules/prefer_conditional_expressions.dart
+++ b/lib/src/rules/prefer_conditional_expressions.dart
@@ -44,19 +44,19 @@ class _Visitor extends RecursiveAstVisitor<void> {
     super.visitIfStatement(node);
 
     if (node.elseStatement != null && node.elseStatement is! IfStatement) {
-      if (_hasAssignment(node) || _hasReturn(node)) {
+      if (_hasBothAssignment(node) || _hasBothReturn(node)) {
         _statements.add(node);
       }
     }
   }
 
-  bool _hasAssignment(IfStatement statement) {
+  bool _hasBothAssignment(IfStatement statement) {
     final thenAssignment = _getAssignmentExpression(statement.thenStatement);
     final elseAssignment = _getAssignmentExpression(statement.elseStatement);
 
     return thenAssignment != null &&
         elseAssignment != null &&
-        _checkAssignmentExpression(thenAssignment, elseAssignment);
+        _haveEqualNames(thenAssignment, elseAssignment);
   }
 
   AssignmentExpression _getAssignmentExpression(Statement statement) {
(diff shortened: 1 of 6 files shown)


lib/src/rules/prefer_conditional_expressions.dart
```diff
@@ -0,0 +1,111 @@
...
+          (elseAssignment.leftHandSide as Identifier).name;
+
+  bool _hasReturn(IfStatement statement) {
+    final thenReturn = _getReturnStatement(statement.thenStatement);
+    final elseReturn = _getReturnStatement(statement.elseStatement);
+
+    return thenReturn != null && elseReturn != null;
+  }
+
+  ReturnStatement _getReturnStatement(Statement statement) {
```
Anrock, line 96: Why return `ReturnStatement` instead of bool if the only place that uses this function only cares about result being not null? In this case function can be as simple as
```dart
(Statement s) => s is ReturnStatement
              || s is Block && s.statements.length == 1 && s.statements.first is ReturnStatement
```

lib/src/rules/prefer_conditional_expressions.dart
```diff
@@ -0,0 +1,111 @@
...
+  AssignmentExpression _getAssignmentExpression(Statement statement) {
+    if (statement is ExpressionStatement &&
+        statement.expression is AssignmentExpression) {
+      return statement.expression as AssignmentExpression;
+    }
+
+    if (statement is Block && statement.statements.length == 1) {
+      final innerStatement = statement.statements.single;
+
+      if (innerStatement is ExpressionStatement &&
```
Anrock, line 71: This piece is exact copy of piece in function beginning. How about:
```dart
AssignmentExpression _getAssignmentExpression(Statement statement) {
  if (statement is ExpressionStatement && statement.expression is AssignmentExpression) {
      return statement.expression as AssignmentExpression;
  }

  if (statement is Block && statement.statements.length == 1) {
      return _getAssignmentExpression(statement.statements.single);
  }

  return null;
}
```

incendial: Yes, but the rule then will trigger on
```dart
if (a == 0) {
    {
      a = 1;
    }
  } else {
    a = 2;
  }
```
And it, actually, seems ok.  I will apply those changes.

lib/src/rules/prefer_conditional_expressions.dart
incendial: I was thinking about similarity between `_getReturnStatement` and `_getAssignmentExpression`, but yeah, your suggestion looks great!

incendial:
@dkrutskikh @Anrock have updated the PR, take a look please.

incendial:
@dkrutskikh do we still need to update Changelog?

dkrutskikh:
@incendial please update changelog

PR #83, opened 2020-07-30
Improve no magic number

### LINKED ISSUE #69: Improve no-magic-number rule
Some field cases with false-positivish warnings:
```dart
final someDay = DateTime(2006, 12, 1); // DateTime has no const constructor, so there is always a warning or each int should be defined as const separately
```
```dart
Future.delayed(const Duration(seconds: 5)); // Triggered despite const constructor.
...
const delay = Duration(seconds: 5));
Future.delayed(delay); // This is fine. Not sure if should be fixed
```
```dart
... Intl.message(..., example: const <String, int>{ 'Assigneed': 3 } // Warning
```

So far it looks like rule should be altered to not trigger if magic number is inside const constructors and number literals in other const literals.
`DateTime` should be made an exception, since it doesn't have a const constructor and putting every argument in separate const declaration is inconvenient.

Anrock:
Resolves #69

### DIFF
--- a/lib/src/rules/no_magic_number_rule.dart
+++ b/lib/src/rules/no_magic_number_rule.dart
@@ -31,10 +31,7 @@ class NoMagicNumberRule extends BaseRule {
 
     return visitor.literals
         .where(_isMagicNumber)
-        .where((lit) =>
-            lit.thisOrAncestorMatching((ancestor) =>
-                ancestor is VariableDeclaration && ancestor.isConst) ==
-            null)
+        .where(_isNotInsideNamedConstant)
         .map((lit) => createIssue(this, _warningMessage, null, null, sourceUrl,
             sourceContent, unit.lineInfo, lit))
         .toList(growable: false);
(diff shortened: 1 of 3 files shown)


lib/src/rules/no_magic_number_rule.dart
```diff
@@ -12,11 +12,11 @@ class NoMagicNumberRule extends BaseRule {
   static const _warningMessage =
       'Avoid using magic numbers. Extract them to named constants';
 
-  // TODO(shtepin): allow configuration of allowed values
-  static const allowedMagicNumbers = [-1, 0, 1];
+  final List<num> _allowedMagicNumbers;
```
dkrutskikh, line 15: @Anrock can you change List on Iterable?

Anrock: @dkrutskikh sure, but what is the point? It's private anyway. For immutability?

dkrutskikh: @Anrock exactly, for immutability.

PR #84, opened 2020-07-30
Allow number of methods warning level configuration

csepanda:
AnalysisOptions missed "number of methods" option reading from configuration.


csepanda:
@dkrutskikh I have added missing config params checks to unit tests of AnalysisOptions.from

Anrock:
I guess we should add number-of-methods option to configuration example in README too.

PR #85, opened 2020-07-30
Add prefer-on-push-cd-strategy

### LINKED ISSUE #81: Use OnPush change detection strategy by default

csepanda:
Closes #81 

**What**
Create validation rule for mandatory use of OnPush CD strategy.

**Why**
OnPush strategy should be used as default because default strategy usage leads to performance issues.


lib/src/rules_factory.dart
```diff
@@ -1,3 +1,5 @@
+import 'package:dart_code_metrics/src/rules/prefer_on_push_cd_strategy.dart';
```
dkrutskikh, line 1: @csepanda in this project, we use relative imports

csepanda: Fixed

Anrock:
Hm, what if we make this rule more generic by making it configurable in preferred CD strategy? In case somebody likes default strategy.

lib/src/rules/prefer_on_push_cd_strategy.dart
```diff
@@ -0,0 +1,75 @@
...
+    final strategyValue = changeDetectionArg.expression;
+    if (strategyValue is! PrefixedIdentifier) {
+      return _expression.add(changeDetectionArg);
+    }
+
+    if (_isStrategyValueCorrect(strategyValue as PrefixedIdentifier)) {
+      return;
+    }
+
+    return _expression.add(changeDetectionArg);
```
incendial, line 64: It seems that you can simplify check and get rid of lines 56-58
```dart
    if (strategyValue is PrefixedIdentifier && _isStrategyValueCorrect(strategyValue)  {
       return; 
    }
    
    return _expression.add(changeDetectionArg);
```

csepanda: indeed, thank you! Fixed.

incendial: I guess you should just `return` instead of `return _expression.add(changeDetectionArg);` on [line 57](https://github.com/wrike/dart-code-metrics/pull/85/files#diff-059454b75e7f06023b6f4ceed0431748R57)

csepanda: Yes, tests are failed. Fixed it. My bad

PR #100, opened 2020-08-06
add Error code issue severity

gerrkoff:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[X] Add something to the core
[ ] Other, please explain:

Add new severity type Error to existing Style & Warning


Anrock:
Nothing wrong with that, however nothing useful too. Do we have error rules? No. Will we have error rules? I doubt that. Also "bug risk" sounds a lot like a "warning" to me, which already exists. All in all I see no sense in merging a priori unused code. Would be fine if it came along with useful rule that uses it.

gerrkoff:
The idea is to give lib consumer an opportunity to override severity of any rule to error and therefore make CI jobs fail.
In fact, I wanted to have something like this 
```
dart_code_metrics:
  rules:
    - prefer-on-push-cd-strategy:
        severity: error
```
as it looks a bit strange to fail a build because of warning. But I'm not insisting :)

Anrock:
@gerrkoff Oh, I see now. In that case yeah, let's have this.

PR #101, opened 2020-08-07
activate some linter rules

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: activate some linter rules and fix found issues

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -59,16 +59,23 @@ linter:
     - control_flow_in_finally
     - diagnostic_describe_all_properties
     - directives_ordering
+    - do_not_use_environment
     - empty_statements
+    - exhaustive_cases
     - file_names
     - flutter_style_todos
     - hash_and_equals
     - invariant_booleans
     - iterable_contains_unrelated_type
     - join_return_with_assignment
+    - leading_newlines_in_multiline_strings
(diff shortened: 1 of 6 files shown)


analysis_options.yaml
```diff
@@ -98,21 +106,29 @@ linter:
     - prefer_typing_uninitialized_variables
     - prefer_void_to_null
     - provide_deprecation_message
+    - sized_box_for_whitespace
```
incendial, line 109: It's a rule for Flutter widgets only, do we really need it?

analysis_options.yaml
```diff
@@ -98,21 +106,29 @@ linter:
...
     - unnecessary_overrides
     - unnecessary_parenthesis
+    - unnecessary_raw_strings
     - unnecessary_statements
+    - unnecessary_string_escapes
+    - unnecessary_string_interpolations
     - unsafe_html
     - use_full_hex_values_for_flutter_colors
+    - use_is_even_rather_than_modulo
+    - use_key_in_widget_constructors
```
incendial, line 130: And this one is only for Flutter too.

PR #104, opened 2020-08-08
Add provided args check for intl message

### LINKED ISSUE #70: Improve prefer-intl-name
Check `args` for message with arguments

incendial:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[x] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

The rule now checks for message arguments and warns when they are not provided to `args`.
Closes #70

### DIFF
--- a/test/rules/prefer_intl_args_test.dart
+++ b/test/rules/prefer_intl_args_test.dart
@@ -1,9 +1,12 @@
 @TestOn('vm')
+
 import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/analysis/utilities.dart';
 import 'package:dart_code_metrics/src/models/code_issue_severity.dart';
 import 'package:dart_code_metrics/src/rules/prefer_intl_name.dart';
 import 'package:test/test.dart';
+
+// ignore_for_file: use_raw_strings
 
 const _incorrectContent = '''
 import 'package:intl/intl.dart';
@@ -42,7 +45,7 @@ class ClassI18n {
 }
 ''';
 
-final _correctContent = '''
+const _correctContent = '''
 import 'package:intl/intl.dart';
 
 final Iterable<String> titles = ['first', 'second'];
--- a/test/rules/prefer_intl_args_test.dart
+++ b/test/rules/prefer_intl_args_test.dart
@@ -1,0 +1,154 @@
+@TestOn('vm')
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/analysis/utilities.dart';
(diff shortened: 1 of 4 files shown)


test/rules/prefer_intl_args_test.dart
```diff
@@ -0,0 +1,157 @@
...
+
+const _incorrectContent = '''
+import 'package:intl/intl.dart';
+
+final Iterable<String> titles = ['first', 'second'];
+final title = 'first';
+
+class ClassI18n {
+  static final String staticFinalFieldInClassTitle = Intl.message(
+    'static final field in class \$title,
```
andrewst, line 19: finish char quote is absent

andrewst, line 19: finish char quote is absent

test/rules/prefer_intl_args_test.dart
```diff
@@ -0,0 +1,157 @@
...
+  name: 'methodExpressionTitleEmptyArgument',
+  args: <Object>[title],
+);
+
+String methodExpressionTitleEmptyArgs() => Intl.message(
+  'method with expression',
+  name: 'methodExpressionTitleEmptyArgs',
+  args: <Object>[],
+);
+''';
```
andrewst, line 83: many cases is not correct because:
`The purpose of wrapping the message in a function is to allow it to
have parameters which can be used in the result. The message string is
allowed to use a restricted form of Dart string interpolation, where
only the function's parameters can be used, and only in simple
expressions. Local variables cannot be used, and neither can
expressions with curly braces. Only the message string can have
interpolation. The name, desc, args, and examples must be literals and
not contain interpolations. Only the args parameter can refer to
variables, and it should list exactly the function parameters. If you
are passing numbers or dates and you want them formatted, you must do
the formatting outside the function and pass the formatted string into
the message.`

https://github.com/dart-lang/intl/blame/master/README.md#L107

incendial: Hm, interesting, makes easier to check then

incendial: @dkrutskikh should we add any warning for that kind of cases?

dkrutskikh: @incendial Why not? I think it's a good idea!

andrewst: I have some work on this rule, I'm ready to make a request today

incendial:
Closed due to #105

PR #110, opened 2020-08-15
Component annotation arguments ordering rule implementation

### LINKED ISSUE #92: [New rule] Component annotation arguments ordering
Enforces ordering for `@Component` Angular annotation arguments. Order should be configurable.
Available options:
* selector
* template(includes templateUrl and template)
* directives (includes directives and directiveTypes)
* pipes
* encapsulation
* styles (includes styles and styleUrls)
* exports (includes exports and exportAs)
* changeDetection
* visibility
* providers (includes providers and viewProviders)

[x] Other (please specify: enforces specific order)

With given config:
```yaml
- selector
- providers
- change-detection
```

Bad:
```dart
@Component(
  changeDetection: ChangeDetectionStrategy.OnPush,
  selector: 'component-selector',
)
```

Good:
```dart
@Component(
  selector: 'component-selector',
  changeDetection: ChangeDetectionStrategy.OnPush,
)
```

Yes

incendial:

[ ] Documentation update
[ ] Bug fix
[x] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Closes #92

### DIFF
--- a/lib/src/rules/component_annotation_arguments_ordering.dart
+++ b/lib/src/rules/component_annotation_arguments_ordering.dart
@@ -1,0 +1,233 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:dart_code_metrics/src/models/code_issue.dart';
+import 'package:dart_code_metrics/src/models/code_issue_severity.dart';
+import 'package:meta/meta.dart';
+
+import 'base_rule.dart';
+import 'rule_utils.dart';
+
+class ComponentAnnotationArgumentsOrderingRule extends BaseRule {
+  static const ruleId = 'component-annotation-arguments-ordering';
+  static const _documentationUrl = 'https://git.io/JJ5HC';
+
+  static const _warningMessage = 'should be before';
+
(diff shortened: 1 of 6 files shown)


lib/src/rules/component_annotation_arguments_ordering.dart
```diff
@@ -0,0 +1,233 @@
...
+  static const exports = _ArgumentGroup._(
+    'exports',
+    ['exports', 'exportAs'],
+  );
+  static const changeDetection = _ArgumentGroup._(
+    'change_detection',
+    ['changeDetection'],
+  );
+
+  static const _groupsOrder = [
```
dkrutskikh, line 184: @incendial I think it's a not good idea have two _groupsOrder. (first int class ComponentAnnotationArgumentsOrderingRule and in global scope)

incendial: @dkrutskikh why not? First one in `ComponentAnnotationArgumentsOrderingRule` is private, and can't be accessed outside the class. Second one is available only through `_ArgumentGroup` class (which is private). Could u please explain more?

PR #121, opened 2020-08-28
Long method

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: add anti pattern long-method

### DIFF
--- a/doc/anti-patterns/long-method.md
+++ b/doc/anti-patterns/long-method.md
@@ -1,0 +1,13 @@
+# Long Method
+
+Long methods are difficult to reuse and understand because they are usually responsible for more than one thing. Separating those methods to several short ones with proper names helps you to reuse your code and understand it better without reading methods body.
+
+Lines of code with clarification comments usually a sign for possible method extraction because you can name extracted method in a way that will cover comment description and then remove comment. Even comments for one line is a sign for method extraction.
+
+* To shorten a method, just apply **Extract Method** command.
+* If local variables and parameters prevent a method extraction, apply **Replace Temp with Query**, **Introduce Parameter Object** or **Preserve Whole Object** commands.
+* Conditional statements or loops indicate the possibility of method extraction. Use **Decompose Conditional** command for conditional expressions and **Extract Method** for loops.
+
+***
+
+P.S. We use terminology from a book **Refactoring Improving the Design of Existing Code** by *Martin Fowler*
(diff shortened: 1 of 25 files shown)


lib/src/reporters/console_reporter.dart
```diff
@@ -35,12 +35,15 @@ class ConsoleReporter implements Reporter {
...
 
-  final _severityHumanReadable = {
+  static const _severityHumanReadable = {
     CodeIssueSeverity.style: 'Style',
     CodeIssueSeverity.warning: 'Warning',
     CodeIssueSeverity.error: 'Error',
   };
 
+  final _designIssuesColor = AnsiPen()..yellow();
+  static const _designIssues = 'Design';
```
incendial, line 45: Why design issues have separate severity calculation? And what do you think about allowing users to configure severity for design issues?

doc/anti-patterns/long-method.md
```diff
@@ -0,0 +1,13 @@
...
+
+If you feel need to comment on something in code, you need extract this block to a method. It makes sense to allocate even one line into a method if it needs clarification.
+
+* To shorten the method, just apply the **Extract Method**.
+* If local variables and parameters prevent to method allocation, you can apply **Replace Temp with Query**, **Introduce Parameter Object** or **Preserve Whole Object**.
+* Conditional statements or loops indicate the possibility of extract into a separate method. For working with a conditional expressions **Decompose Conditional** is suitable. To work with a loops use **Extract Method**.
+
+***
+
+P.S. We use terminology from book **Refactoring Improving the Design of Existing Code** by *Martin Fowler*
```
incendial, line 13: Long methods are difficult to reuse and understand because they are usually responsible for more than one thing. Separating those methods to several short ones with proper names helps you to reuse your code and understand it better without reading methods body.

Lines of code with clarification comments usually a sign for possible method extraction because you can name extracted method in a way that will cover comment description and then remove comment. Even comments for one line is a sign for method extraction.

* To shorten a method, just apply **Extract Method** command.
* If local variables and parameters prevent a method extraction, apply **Replace Temp with Query**, **Introduce Parameter Object** or **Preserve Whole Object** commands.
* Conditional statements or loops indicate the possibility of method extraction. Use **Decompose Conditional** command for conditional expressions and **Extract Method** for loops.
***

P.S. We use terminology from a book **Refactoring Improving the Design of Existing Code** by *Martin Fowler*.

dkrutskikh: update documetation

PR #130, opened 2020-09-06
Potential null dereference rule implementation

### LINKED ISSUE #111: [New rule] Potential null dereference
Check for potential null dereference in conditional expressions and blocks.

https://www.viva64.com/en/w/v6008/

[ ] Other (please specify:)

```dart
if (obj != null || obj.foo()) { ... } // in obj.foo invocation obj could be null

if (obj == null && obj.foo()) { ... } // in obj.foo invocation obj is always null

if (obj == null) {
  ...
  var bar = obj.foo(); // in obj.foo invocation obj is always null
  ...
}
```

Yes

incendial:

[ ] Documentation update
[ ] Bug fix
[X] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Closes #111


lib/src/rules_factory.dart
```diff
@@ -1,3 +1,5 @@
+import 'package:dart_code_metrics/src/rules/potential_null_dereference.dart';
```
dkrutskikh, line 1: @incendial please use relative import

incendial: Fixed

lib/src/rules/potential_null_dereference.dart
```diff
@@ -0,0 +1,190 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/syntactic_entity.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:dart_code_metrics/src/models/code_issue.dart';
```
dkrutskikh, line 5: @incendial please use relative import

PR #150, opened 2020-11-04
Explain how to use the tool with Flutter

creativecreatorormaybenot:

- [x] Documentation update

I explained how to use the tool with Flutter. This was not obvious before for two reasons:

* When using Flutter, the Pub executables directory is not added to path by default.
* The `metrics` executable uses `pub`, which is not available when using Flutter (only `flutter pub` is available).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,13 +69,13 @@ pub global activate dart_code_metrics
 metrics lib
 ```
 
-If you want to run the tool directly (e.g. because you are using Flutter), you can use:
+#### Flutter usage
 
 ```bash
 flutter pub global run dart_code_metrics:metrics lib
 ```
 
-Full usage:
+#### Full usage:
 
 ```text
 Usage: metrics [options...] <directories>
--- a/README.md
+++ b/README.md
@@ -69,6 +69,12 @@ pub global activate dart_code_metrics
 metrics lib
 ```
 
+If you want to run the tool directly (e.g. because you are using Flutter), you can use:
+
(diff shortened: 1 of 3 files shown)


creativecreatorormaybenot:
(Checks failing is not due to this PR)

incendial:
@creativecreatorormaybenot Hi! Thank you for contribution. Unfortunately, we have lack of experience with Flutter, so if you have any other feedback, we would really appreciate it.
I will take a look why checks are failing.

README.md
```diff
@@ -69,6 +69,12 @@ pub global activate dart_code_metrics
 metrics lib
 ```
 
+If you want to run the tool directly (e.g. because you are using Flutter), you can use:
```
incendial, line 72: @creativecreatorormaybenot because it's `flutter pub` command, it seems to me that we could rename this block to `Flutter usage:`. What do you think?

incendial (changes requested):

README.md
creativecreatorormaybenot: @incendial You are right, I missed that.

creativecreatorormaybenot:
@incendial Thanks, I am just starting out to use it with Flutter.

I tested it on one repo and it worked. With some other repos (maybe also with null safety), I will have to see. I will let you know when I find out more 👍

incendial:
@creativecreatorormaybenot LGTM, thanks again for contribution.

PR #153, opened 2020-11-07
Prefer trailing comma rule implementation

### LINKED ISSUE #152: [New rule] Prefer trailing comma
Should check for trailing comma for arguments, parameters, enum values and collections. 
By default should warn in cases when items aren't on a single line.
Additionally should accept items count breakpoint config. For example, if breakpoint count is equal 2, then the rule should warn when items count is greater than or equal to the count or items aren't on a single line.

[X] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

Without config:
Bad:
```dart
void firstFunction(String arg1, String arg2,
  String arg3) {
  return;
}
```
Good:

```dart
void firstFunction(
  String arg1,
  String arg2,
  String arg3,
) {
  return;
}
```

With given config:
```
- break_on: 2
```
Bad:
```dart
void firstFunction(String arg1, String arg2,
  String arg3) {
  return;
}

void secondFunction(String arg1, String arg2, String arg3) {
  return;
}
```
Good:

```dart
void firstFunction(
  String arg1,
  String arg2,
  String arg3,
) {
  return;
}

void secondFunction(
  String arg1,
  String arg2,
  String arg3,
) {
  return;
}

enum FirstEnum { firstItem }

const instance = FirstClass(
  0,
  0,
  0,
);
```
Yes

incendial:

[ ] Documentation update
[ ] Bug fix
[X] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Closes #152

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -2,6 +2,7 @@ include: package:pedantic/analysis_options.yaml
 
 analyzer:
   exclude:
+    - test/**/examples/**
     - example/**
     - test/resources/**
   plugins:
@@ -34,10 +35,10 @@ linter:
   rules:
     - always_put_control_body_on_new_line
     - always_put_required_named_parameters_first
-#    - always_specify_types
-#    - always_use_package_imports
+    #    - always_specify_types
+    #    - always_use_package_imports
     - avoid_annotating_with_dynamic
-#    - avoid_as
+    #    - avoid_as
     - avoid_bool_literals_in_conditional_expressions
     - avoid_catches_without_on_clauses
     - avoid_catching_errors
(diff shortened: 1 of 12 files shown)


analysis_options.yaml
```diff
@@ -34,10 +35,10 @@ linter:
   rules:
     - always_put_control_body_on_new_line
     - always_put_required_named_parameters_first
-#    - always_specify_types
-#    - always_use_package_imports
+    #    - always_specify_types
```
dkrutskikh, line 38: @incendial please revert this strange changes

dkrutskikh (changes requested):

Anrock (approved):

PR #163, opened 2020-11-21
Prefer entry point imports rule implementation

### LINKED ISSUE #162: [New rule] Prefer entry point imports
Should check that libraries in `lib/src` are imported from entry point exports in the same package.

[ ] Other (please specify:)

In a file outside lib.

Bad:
```dart
import 'package:package_name/src/implementation/export.dart
```

Good:
```dart
import 'package:package_name/export.dart
```

Yes

incendial:

[ ] Documentation update
[ ] Bug fix
[X] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Closes #162

### DIFF
--- a/test/rules/prefer_entry_point_imports/prefer_entry_point_imports_test.dart
+++ b/test/rules/prefer_entry_point_imports/prefer_entry_point_imports_test.dart
@@ -1,0 +1,83 @@
+import 'dart:io';
+
+@TestOn('vm')
+import 'package:analyzer/dart/analysis/utilities.dart';
+import 'package:dart_code_metrics/src/models/code_issue_severity.dart';
+import 'package:dart_code_metrics/src/models/source.dart';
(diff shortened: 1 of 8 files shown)


test/rules/prefer_entry_point_imports/prefer_entry_point_imports_test.dart
```diff
@@ -0,0 +1,83 @@
+import 'dart:io';
+
+@TestOn('vm')
```
dkrutskikh, line 3: @incendial annotation must be on first line

incendial: Fixed

Anrock (approved):

PR #181, opened 2020-12-25
Avoid unused parameters rule implementation

### LINKED ISSUE #161: [New rule] Avoid unused parameters
Should check for unused parameters in function or method body.

[ ] Other (please specify:)

Bad:
```dart
void someFunction(String s) {
  return;
}

class SomeClass {
  void method(String s) {
    return;
  }
}
```

Good:
```dart
void someFunction() {
  return;
}

class SomeClass {
  void method() {
    return;
  }
}

void someFunction(String s) {
  print(s);
  return;
}

class SomeClass {
  void method(String s) {
    print(s);
    return;
  }
}
```

Yes

incendial:

[ ] Documentation update
[ ] Bug fix
[X] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Closes #161


doc/rules/avoid_unused_parameters.md
```diff
@@ -0,0 +1,54 @@
...
+  void method(String s) {
+    return;
+  }
+}
+```
+
+Good:
+
+```dart
+void someFunction() {
```
Anrock, line 33: I guess you can add a comment here about that rule isn't triggered because there is no unused parameter, not because it was deleted as a fix from `Bad` example.

doc/rules/avoid_unused_parameters.md
```diff
@@ -0,0 +1,54 @@
...
+    return;
+  }
+}
+
+void someFunction(String s) {
+  print(s);
+  return;
+}
+
+class SomeClass {
```
Anrock, line 48: Add some example with unused params renamed to `_` as mentioned in description?

doc/rules/avoid_unused_parameters.md
incendial: Hmm... does it really matter? Will it be clear if i just rename function to `someOtherFunction`?

doc/rules/avoid_unused_parameters.md
incendial: Yeah, good idea, i've updated the docs

doc/rules/avoid_unused_parameters.md
Anrock: It's just at first I read the "good" example as the "fixed" version. Renaming the function should help clear it up.

Anrock (approved):

PR #196, opened 2021-01-03
Weight Of Class metric

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/doc/metrics/weight-of-class.md
+++ b/doc/metrics/weight-of-class.md
@@ -1,0 +1,7 @@
+# Weight Of Class
+
+Number of **functional** public methods divided by the total number of public methods.
+
+This metric tries to quantify whether the measured class' interface reveals more data than behaviour. Low values (less than 33%) indicate that the class reveals much more data than behaviour, which is a sign of poor encapsulation.
+
+Our definition of **functional method** excludes getters and setters.
--- a/lib/src/config/cli/arguments_parser.dart
+++ b/lib/src/config/cli/arguments_parser.dart
@@ -64,10 +64,10 @@ void _appendReporterOption(ArgParser parser) {
 }
 
 @immutable
-class _MetricOption {
+class _MetricOption<T> {
   final String name;
   final String help;
-  final int defaultValue;
+  final T defaultValue;
 
   const _MetricOption(this.name, this.help, this.defaultValue);
 }
@@ -99,6 +99,11 @@ void _appendMetricsThresholdOptions(ArgParser parser) {
       'Maximum nesting threshold',
       maximumNestingDefaultWarningLevel,
     ),
+    _MetricOption(
+      weightOfClassKey,
+      'Weight Of Class threshold',
+      weightOfClassDefaultWarningLevel,
+    ),
   ];
 
   for (final metric in metrics) {
@@ -107,9 +112,16 @@ void _appendMetricsThresholdOptions(ArgParser parser) {
       help: metric.help,
       valueHelp: '${metric.defaultValue}',
       // ignore: avoid_types_on_closure_parameters
-      callback: (String i) {
-        if (i != null && int.tryParse(i) == null) {
-          print("'$i' invalid value for argument ${metric.name}");
+      callback: (String value) {
+        var invalid = true;
+        if (metric.defaultValue is int) {
+          invalid = value != null && int.tryParse(value) == null;
+        } else if (metric.defaultValue is double) {
+          invalid = value != null && double.tryParse(value) == null;
+        }
+
+        if (invalid) {
+          print("'$value' invalid value for argument ${metric.name}");
         }
       },
     );
--- a/lib/src/metrics_analyzer.dart
+++ b/lib/src/metrics_analyzer.dart
@@ -78,6 +78,8 @@ class MetricsAnalyzer {
             .map((entity) => entity.path))
         .toList();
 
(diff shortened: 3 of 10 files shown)


lib/src/metrics/weight-of-class/weight_of_class_metric.dart
```diff
@@ -0,0 +1,38 @@
...
+        .toList();
+
+    final functionalMethods = totalPublicMethods.where((function) {
+      if (function.type == FunctionType.constructor) {
+        return false;
+      }
+
+      final declaration = function.declaration;
+      if (declaration is MethodDeclaration) {
+        return !declaration.isGetter && !declaration.isGetter;
```
incendial, line 24: Looks like there should be `setter` check

lib/src/metrics/weight-of-class/weight_of_class_metric.dart
```diff
@@ -0,0 +1,38 @@
...
+import 'package:code_checker/checker.dart';
+
+import '../../utils/declaration_utils.dart';
+
+class WeightOfClassMetric {
+  double compute(
+    ScopedClassDeclaration component,
+    Iterable<ScopedFunctionDeclaration> functions,
+  ) {
+    final totalPublicMethods = componentFunctions(component, functions)
```
incendial, line 11: Seems that u should exclude constructor earlier, because `totalPublicMethods` will always greater than `functionalMethods`

lib/src/metrics/weight-of-class/weight_of_class_metric.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:code_checker/checker.dart';
+
+import '../../utils/declaration_utils.dart';
+
+class WeightOfClassMetric {
+  double compute(
+    ScopedClassDeclaration component,
```
incendial, line 8: Why classes are called components?

lib/src/metrics_analyzer.dart
```diff
@@ -125,6 +128,7 @@ class MetricsAnalyzer {
                     .where((function) =>
                         function.enclosingDeclaration == component)
                     .length,
+                weightOfClass: woc.compute(component, visitor.functions),
```
incendial, line 131: Same as https://github.com/wrike/dart-code-metrics/pull/196/files#r550990857

lib/src/config/cli/arguments_parser.dart
```diff
@@ -108,7 +113,14 @@ void _appendMetricsThresholdOptions(ArgParser parser) {
       valueHelp: '${metric.defaultValue}',
       // ignore: avoid_types_on_closure_parameters
       callback: (String i) {
```
incendial, line 115: Could `i` be renamed to something more meaningful?

doc/metrics/weight-of-class.md
```diff
@@ -0,0 +1,7 @@
+# Weight Of Class
+
+Number of **functional** public methods divided by the total number of public methods.
+
+This metric tries to quantify whether the measured class' interface reveals more data than behaviour. Low values (less than 33%) indicate that the class reveals much more data than behaviour, which is a sign of poor encapsulation.
+
+Our definition of **functional method** excludes getters and setters.
```
incendial, line 7: From my point of view this metric is very strange: so basically we measure amount of methods without getters and setters, but if a class will have public fields instead of getters/setters this metric will not be low (but basically, idea is the same - more data than behaviour.

incendial (changes requested):
@dkrutskikh could u take a look on my comments? Also, couldn't find tests for the weightOfClass value calculation

lib/src/metrics/weight-of-class/weight_of_class_metric.dart
dkrutskikh: fix

lib/src/metrics/weight-of-class/weight_of_class_metric.dart
dkrutskikh: fix

lib/src/config/cli/arguments_parser.dart
dkrutskikh: fix

lib/src/metrics/weight-of-class/weight_of_class_metric.dart
incendial: how should we handle other 'component' references in the codebase? different request?

dkrutskikh: exactly

dkrutskikh:
@incendial I've added tests

lib/src/metrics/weight-of-class/weight_of_class_metric.dart
incendial: still component reference https://github.com/wrike/dart-code-metrics/pull/196/files#diff-640bb8695d3503dd75b8fff7609d52ce2ff936ac4b9fffe67342158fee786059R35

incendial: could u please create an issue for renaming then?

PR #229, opened 2021-03-21
added suppression

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/models/function_type.dart
+++ b/lib/src/models/function_type.dart
@@ -2,19 +2,19 @@
 ///
 /// Used when reporting.
 class FunctionType {
-  /// The class entity representing a class constructor.
+  /// The function entity representing a class constructor.
   static const constructor = FunctionType._('constructor');
 
-  /// The class entity representing a class method.
+  /// The function entity representing a class method.
   static const method = FunctionType._('method');
 
-  /// The class entity representing a function.
+  /// The function entity representing a generic function.
   static const function = FunctionType._('function');
 
-  /// The class entity representing a getter.
+  /// The function entity representing a getter.
   static const getter = FunctionType._('getter');
 
-  /// The class entity representing a setter.
+  /// The function entity representing a setter.
   static const setter = FunctionType._('setter');
 
   final String _value;
--- a/lib/src/models/function_type.dart
+++ b/lib/src/models/function_type.dart
@@ -1,0 +1,26 @@
+/// Enum class for type of a function entity.
+///
+/// Used when reporting.
+class FunctionType {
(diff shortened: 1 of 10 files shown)


lib/src/models/function_type.dart
```diff
@@ -0,0 +1,26 @@
...
+/// Used when reporting.
+class FunctionType {
+  /// The class entity representing a class constructor.
+  static const constructor = FunctionType._('constructor');
+
+  /// The class entity representing a class method.
+  static const method = FunctionType._('method');
+
+  /// The class entity representing a function.
+  static const function = FunctionType._('function');
```
incendial, line 12: I think the comment should not mention a class then if we talk about a function?

lib/src/suppression.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:analyzer/source/line_info.dart';
+
+/// Represents information about rule suppression
```
incendial, line 3: Correction:  `Represents an information about rule suppression`

lib/src/suppression.dart
```diff
@@ -0,0 +1,48 @@
...
+class Suppression {
+  static final _ignoreMatchers = RegExp('//[ ]*ignore:(.*)', multiLine: true);
+
+  static final _ignoreForFileMatcher =
+      RegExp('//[ ]*ignore_for_file:(.*)', multiLine: true);
+
+  final _ignoreMap = <int, List<String>>{};
+  final _ignoreForFileSet = <String>{};
+
+  /// Checks [id] globally suppressed
```
incendial, line 13: Correction: `Checks that the [id] is globally suppressed`

lib/src/suppression.dart
```diff
@@ -0,0 +1,48 @@
...
+  bool isSuppressed(String id) => _ignoreForFileSet.contains(_canonicalize(id));
+
+  /// Checks [id] suppressed for [lineIndex]
+  bool isSuppressedAt(String id, int lineIndex) =>
```
incendial, line 17: Correction: `Checks that the [id] is suppressed for the [lineIndex]`

incendial (changes requested):

lib/src/models/function_type.dart
incendial: Or is it about an inner function inside of a method?

incendial (approved):

PR #230, opened 2021-03-21
added scope visitor

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/models/scoped_class_declaration.dart
+++ b/lib/src/models/scoped_class_declaration.dart
@@ -1,0 +1,28 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:meta/meta.dart';
+
+import 'class_type.dart';
+
+/// Represents a declaration of a class / mixin / extension.
+@immutable
+class ScopedClassDeclaration {
+  /// Type of declared class entity.
+  final ClassType type;
+
+  /// The node that represents in the AST structure for a Dart program.
+  final CompilationUnitMember declaration;
+
+  /// Returns user defined name.
+  String get name {
+    if (declaration is ExtensionDeclaration) {
+      return (declaration as ExtensionDeclaration).name.name;
+    } else if (declaration is NamedCompilationUnitMember) {
+      return (declaration as NamedCompilationUnitMember).name.name;
+    }
+
+    return '';
+  }
+
+  /// Initialize a newly created [ScopedClassDeclaration] with the given [type] and [declaration].
+  const ScopedClassDeclaration(this.type, this.declaration);
+}
--- a/lib/src/models/scoped_class_declaration.dart
+++ b/lib/src/models/scoped_class_declaration.dart
@@ -9,18 +9,21 @@ class ScopedClassDeclaration {
   /// The type of the declared class entity.
   final ClassType type;
 
-  /// The node that represents in the AST structure for a Dart program.
+  /// The node that represents in the AST structure a Dart code snippet.
   final CompilationUnitMember declaration;
 
   /// Returns the user defined name.
   String get name {
-    if (declaration is ExtensionDeclaration) {
-      return (declaration as ExtensionDeclaration).name.name;
-    } else if (declaration is NamedCompilationUnitMember) {
-      return (declaration as NamedCompilationUnitMember).name.name;
+    final node = declaration;
+    String name;
+
+    if (node is ExtensionDeclaration) {
+      name = node.name.name;
+    } else if (node is NamedCompilationUnitMember) {
+      name = node.name.name;
     }
 
-    return '';
+    return name ?? '';
   }
 
   /// Initialize a newly created [ScopedClassDeclaration] with the given [type] and [declaration].
--- a/lib/src/models/scoped_class_declaration.dart
+++ b/lib/src/models/scoped_class_declaration.dart
@@ -6,13 +6,13 @@ import 'class_type.dart';
 /// Represents a declaration of a class / mixin / extension.
 @immutable
 class ScopedClassDeclaration {
-  /// Type of declared class entity.
+  /// The type of the declared class entity.
   final ClassType type;
 
   /// The node that represents in the AST structure for a Dart program.
   final CompilationUnitMember declaration;
 
-  /// Returns user defined name.
+  /// Returns the user defined name.
   String get name {
     if (declaration is ExtensionDeclaration) {
       return (declaration as ExtensionDeclaration).name.name;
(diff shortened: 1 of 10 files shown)


lib/src/models/scoped_class_declaration.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:meta/meta.dart';
+
+import 'class_type.dart';
+
+/// Represents a declaration of a class / mixin / extension.
+@immutable
+class ScopedClassDeclaration {
+  /// Type of declared class entity.
```
incendial, line 9: Correction: `The type of the declared class entity`

lib/src/models/scoped_class_declaration.dart
```diff
@@ -0,0 +1,28 @@
...
+  final ClassType type;
+
+  /// The node that represents in the AST structure for a Dart program.
```
incendial, line 12: Seems that there is something missing after `represents` or I didn't catch the meaning...

lib/src/models/scoped_class_declaration.dart
```diff
@@ -0,0 +1,28 @@
...
+  final CompilationUnitMember declaration;
+
+  /// Returns user defined name.
```
incendial, line 15: Correction: `Returns the user defined name`

lib/src/models/scoped_class_declaration.dart
```diff
@@ -0,0 +1,28 @@
...
+  String get name {
+    if (declaration is ExtensionDeclaration) {
```
incendial, line 17: Seems that if you introduce a local variable like `final declaration = this.declaration;` and then type check it, the analyzer will not require a type cast.

lib/src/models/scoped_function_declaration.dart
```diff
@@ -0,0 +1,51 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:meta/meta.dart';
+
+import 'function_type.dart';
+import 'scoped_class_declaration.dart';
+
+/// Represents a declaration of function / method.
+@immutable
+class ScopedFunctionDeclaration {
+  /// Type of declared function entity.
```
incendial, line 10: Correction: `The type of the declared function entity.`

lib/src/models/scoped_function_declaration.dart
```diff
@@ -0,0 +1,51 @@
...
+    if (declaration is FunctionDeclaration) {
+      return (declaration as FunctionDeclaration).name?.name;
+    } else if (declaration is ConstructorDeclaration) {
+      return (declaration as ConstructorDeclaration).name?.name ??
+          (declaration.parent as NamedCompilationUnitMember).name?.name;
+    } else if (declaration is MethodDeclaration) {
+      return (declaration as MethodDeclaration).name?.name;
+    }
+
+    return null;
```
incendial, line 30: Why return null here and empty string [here](https://github.com/dart-code-checker/dart-code-metrics/pull/230/files#diff-d7f66dd0ebfb0a270e96a557740e36f9909320394ff085a6520d10cff38b1d97R23)?

lib/src/scope_visitor.dart
```diff
@@ -0,0 +1,88 @@
...
+  /// Returns the declarations of the found classes.
+  Iterable<ScopedClassDeclaration> get classes => _classes;
+
+  /// Returns the declarations of the found functions.
+  Iterable<ScopedFunctionDeclaration> get functions => _functions;
+
+  @override
+  void visitClassDeclaration(ClassDeclaration node) {
+    _registerClassDeclaration(ClassType.generic, node, () {
+      super.visitClassDeclaration(node);
```
incendial, line 25: Not clear, why do we need to call the base class visit method as a callback?

test/resources/class_with_factory_constructors.dart
```diff
@@ -0,0 +1,22 @@
+class Logger {
+  final String name;
+  bool mute = false;
+
+  // _cache is library-private, thanks to
+  // the _ in front of its name.
```
incendial, line 6: Do we need this comment in tests?

test/scope_visitor_test.dart
```diff
@@ -0,0 +1,184 @@
...
+      final function = visitor.functions.single;
+      expect(function.type, equals(FunctionType.method));
+      expect(function.name, equals('doSomething'));
+      expect(function.fullName, equals('Doer.doSomething'));
+    });
+
+    test('class with factory constructors', () async {
+      (await resolveFile(
+        path: p.normalize(p
+            .absolute('./test/resources/class_with_factory_constructors.dart')),
```
incendial, line 39: Suggestion: what do you think about moving path declaration to a separate variable (same for all test cases)?

lib/src/scope_visitor.dart
```diff
@@ -0,0 +1,88 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+
+import 'models/class_type.dart';
+import 'models/function_type.dart';
+import 'models/scoped_class_declaration.dart';
+import 'models/scoped_function_declaration.dart';
+
+/// Visitor to collect declarations of classes and functions.
```
incendial, line 9: Correction: `A visitor to collect declarations of classes and functions.`

lib/src/models/scoped_function_declaration.dart
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /// Returns full user defined name.
```
incendial, line 33: Correction: `Returns the full user defined name.`

lib/src/models/scoped_function_declaration.dart
```diff
@@ -0,0 +1,51 @@
...
+  /// Type of declared function entity.
+  final FunctionType type;
+
+  /// The node that represents in the AST structure for a Dart program.
+  final Declaration declaration;
+
+  /// The class declaration of the class to which this function belongs.
+  final ScopedClassDeclaration enclosingDeclaration;
+
+  /// Returns user defined name.
```
incendial, line 19: Correction: `Returns the user defined name.`

incendial (changes requested):

test/resources/class_with_factory_constructors.dart
dkrutskikh: Yes. This is good example class for future testing some other cases.

incendial (approved):

PR #233, opened 2021-03-24
Metrics

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:


lib/src/metrics/maximum_nesting_level/maximum_nesting_level_metric.dart
```diff
@@ -0,0 +1,91 @@
...
+            .capitalize();
+
+        return ContextMessage(
+          message: '$message increases depth',
+          location: nodeLocation(node: block, source: source),
+        );
+      }).toList()
+        ..sort((a, b) => a.location.start.compareTo(b.location.start));
+
+  String _removeImpl(String typeName) {
```
incendial, line 84: Looks like a good candidate to be moved to helpers as it also used here https://github.com/dart-code-checker/dart-code-metrics/pull/233/files#diff-1cf40e64be3c42be949ff9051f9ee3c0ebe745a52d2ad2c34fde518288841f12R89

lib/src/metrics/cyclomatic_complexity/cyclomatic_complexity_metric.dart
```diff
@@ -0,0 +1,96 @@
...
+  measuredType: EntityType.methodEntity,
+  examples: [],
+);
+
+/// Cyclomatic Complexity (CYCLO)
+///
+/// Cyclomatic complexity is a measure of the code's complexity achieved by
+/// measuring the number of linearly independent paths through a source code.
+class CyclomaticComplexityMetric extends FunctionMetric<int> {
+  /// Id of this metric.
```
incendial, line 31: As for useless documentation comments, sorry, but that one looks kinda it. Not additional context provided, nothing, just a name of the field in different words.

lib/src/metrics/number_of_methods_metric.dart
```diff
@@ -0,0 +1,80 @@
...
+      context: _context(methods, source),
+    );
+  }
+
+  @override
+  String commentMessage(String nodeType, int value, int threshold) {
+    final methods = '$value ${value == 1 ? 'method' : 'methods'}';
+    final exceeds = value > threshold
+        ? ', which exceeds the maximum of $threshold allowed'
+        : '';
```
incendial, line 59: If we assume that a Metric threshold is always `num`, shouldn't this calculation be moved to the base class then? As I can see, right now it just copy pasted to all implementation classes.

incendial (changes requested):

PR #255, opened 2021-04-08
fix: read analysis_options in plugin

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/obsoleted/analyzer_plugin/analyzer_plugin.dart
+++ b/lib/src/obsoleted/analyzer_plugin/analyzer_plugin.dart
@@ -425,14 +425,18 @@ class MetricsAnalyzerPlugin extends ServerPlugin {
           : null;
 
   Config? _readOptions(AnalysisDriver driver) {
-    final file = driver.analysisContext?.contextRoot.optionsFile;
-    if (file != null && file.exists) {
-      return Config.fromAnalysisOptions(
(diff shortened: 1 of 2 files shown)


lib/src/obsoleted/analyzer_plugin/analyzer_plugin.dart
```diff
@@ -425,14 +425,18 @@ class MetricsAnalyzerPlugin extends ServerPlugin {
...
-      return Config.fromAnalysisOptions(
-        AnalysisOptions(yamlMapToDartMap(
-          AnalysisOptionsProvider(driver.sourceFactory)
-              .getOptionsFromFile(file),
-        )),
-      );
+    // ignore: deprecated_member_use
+    final optionsPath = driver.contextRoot?.optionsFilePath;
+    if (optionsPath?.isNotEmpty ?? false) {
+      final file = resourceProvider.getFile(optionsPath!);
```
incendial, line 431: I'd suggest to do it that way.

```dart
final filePath = driver.contextRoot?.optionsFilePath;
    if (filePath != null && filePath.isNotEmpty) {
      final file = resourceProvider.getFile(filePath);
```

incendial (approved):

PR #258, opened 2021-04-08
Add separate rule `exclude` config

incendial:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[X] Add something to the core
[ ] Other, please explain:


incendial:
@dkrutskikh I actually don't know how to test it correctly, any suggestions?

lib/src/utils/rule_utils.dart
```diff
@@ -25,13 +27,25 @@ Issue createIssue({
...
       pathSegments: ['dart-code-metrics', 'rules', '$ruleId.html'],
     );
 
-/// Returns the [Severity] from map based [config] otherwise [defaultValue]
+/// Returns a [Severity] from map based [config] otherwise [defaultValue]
 Severity readSeverity(Map<String, Object?> config, Severity defaultValue) =>
     Severity.fromString(config['severity'] as String?) ?? defaultValue;
+
+/// Returns a list of excludes from the given [config]
+Iterable<String> readExcludes(Map<String, Object> config) {
```
dkrutskikh, line 42: @incendial please write separate test for this function

lib/src/obsoleted/metrics_analyzer.dart
```diff
@@ -285,8 +290,17 @@ class MetricsAnalyzer {
   Iterable<Issue> _checkOnCodeIssues(
     Suppression ignores,
     InternalResolvedUnitResult source,
+    String filePath,
```
dkrutskikh, line 293: @incendial I'd suggest to receive here absolute path

dkrutskikh (changes requested):

PiN73:
Would it be a good option to exclude some pattern from all dart_code_metrics rules (keeping standard dart rules)?
```yaml
dart_code_metrics:
  exclude:
    - lib/**.freezed.dart
```

incendial:
@PiN73 Could you please elaborate on the motivation for you to exclude only Dart Code Metrics rules? It's definitely not a part of this request, but could be implemented separately.

PiN73:
@incendial yes, it would be useful to allow excluding generated files (like .g.dart or .freezed.dart from [freezed](https://pub.dev/packages/freezed)) because I don't control its contents at all

incendial:
@PiN73 okay, sounds reasonable to me, but I don't think that we'd add this feature in 3.1. Could you please create a separate issue for it with the motivation?

lib/src/obsoleted/metrics_analyzer.dart
incendial: Discussed it, decided to keep as is for now.

lib/src/utils/rule_utils.dart
incendial: Done

dkrutskikh (approved):

PR #264, opened 2021-04-11
Json

dkrutskikh:
### What is the purpose of this pull request?

- [ ] Documentation update
- [ ] Bug fix
- [ ] New rule
- [ ] Add a CLI option
- [ ] Other, please explain:

### DIFF
--- a/doc/reporters/json.md
+++ b/doc/reporters/json.md
@@ -1,0 +1,184 @@
+# JSON reporter
+
+Dart Code Metrics can generate report in JSON format. Use `--reporter=json` to enable this format. A sample command to analyze package and output:
+
+```sh
+dart pub run dart_code_metrics:metrics --reporter=json lib
+
+# or for a Flutter package
+flutter pub run dart_code_metrics:metrics --reporter=json lib
+```
+
+This reporter prints a single JSON object containing some header information, and then the violations grouped by file.
+
+## The **root** object fields are
+
+- `formatVersion` - an integer which will be incremented if we change the serialization format.
+- `timestamp` - a creation time of the report
+- `records` - an array of [objects](#the-record-object-fields-are)
+
+```JSON
+{
+  "formatVersion": 2,
+  "timestamp": "2021-04-11 14:44:42.433491",
+  "records": [
+    {
+      ...
+    },
+    {
+      ...
+    },
+    {
+      ...
+    }
+  ]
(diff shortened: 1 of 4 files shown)


doc/reporters/json.md
```diff
@@ -0,0 +1,184 @@
...
+# or for a Flutter package
+flutter pub run dart_code_metrics:metrics --reporter=json lib
+```
+
+This reporter prints a single JSON object containing some header information, and then the violations grouped by file.
+
+## The **root** object fields are
+
+- `formatVersion` - an integer which will be incremented if we change the serialization format.
+- `timestamp` - a creation time of the report
```
incendial, line 17: Please add info about the DateTime format.

doc/reporters/json.md
```diff
@@ -0,0 +1,184 @@
...
+      ...
+    }
+  ]
+}
+```
+
+## The **record** object fields are
+
+- `path` - a relative path to the target file
+- `classes` - a map with **class name** ad the **key** and **[class report](#the-report-object-fields-are)** as the **value**
```
incendial, line 41: `Correction:` ad -> as

doc/reporters/json.md
```diff
@@ -0,0 +1,184 @@
...
+- `functions` - a map with **function name** ad the **key** and **[function report](#the-report-object-fields-are)** as the **value**
```
incendial, line 42: Same for `ad`.

doc/reporters/json.md
```diff
@@ -0,0 +1,184 @@
...
+  "column": 1
+}
+```
+
+## The **metric value** object fields are
+
+- `metricsId` - an id of the computed metric
+- `value` - an actual value computed by the metric
+- `level` - a level of this value computed by the metric
+- `comment` - a message with information about this value
```
incendial, line 117: `Correction:` this -> the

doc/reporters/json.md
```diff
@@ -0,0 +1,184 @@
...
+- `recommendation` - a message with information about how the user can improve this value *(optional)*
```
incendial, line 118: Same for `this`.

incendial (changes requested):
Please take a look on `this` usages in the `Issue` block. I think we should use `the` instead.

incendial (approved):

PR #270, opened 2021-04-20
fix: skip unanalyzed files

dkrutskikh:

[ ] Documentation update
[x] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Skip files whose not analyze by analyzer package.
Like this:
```
./ios/.symlinks/plugins/**
```

### DIFF
--- a/lib/src/analyzer_plugin/analyzer_plugin.dart
+++ b/lib/src/analyzer_plugin/analyzer_plugin.dart
@@ -174,13 +174,18 @@ class MetricsAnalyzerPlugin extends ServerPlugin {
     ResolvedUnitResult analysisResult,
   ) {
     try {
-      if (analysisResult.unit != null) {
+      final path = analysisResult.path;
+      if (analysisResult.unit != null &&
+          path != null &&
+          (driver.analysisContext?.contextRoot.isAnalyzed(path) ?? false)) {
         final fixes = _check(driver, analysisResult);
 
-        channel.sendNotification(plugin.AnalysisErrorsParams(
-          analysisResult.path!,
-          fixes.map((fix) => fix.error).toList(),
-        ).toNotification());
+        channel.sendNotification(
+          plugin.AnalysisErrorsParams(
+            path,
+            fixes.map((fix) => fix.error).toList(),
+          ).toNotification(),
+        );
       } else {
         channel.sendNotification(
           plugin.AnalysisErrorsParams(analysisResult.path!, [])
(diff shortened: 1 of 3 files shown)


lib/src/obsoleted/analyzer_plugin/analyzer_plugin.dart
```diff
@@ -187,13 +187,18 @@ class MetricsAnalyzerPlugin extends ServerPlugin {
     ResolvedUnitResult analysisResult,
   ) {
     try {
-      if (analysisResult.unit != null) {
+      final path = analysisResult.path;
+      if (analysisResult.unit != null &&
+          path != null &&
+          (driver.analysisContext?.contextRoot.isAnalyzed(path) ?? false)) {
```
incendial, line 193: It seems that we have null `analysisContext` and this won't work (that was the problem with `optionsFilePath`).

incendial (changes requested):

lib/src/analyzer_plugin/analyzer_plugin.dart
```diff
@@ -174,13 +174,18 @@ class MetricsAnalyzerPlugin extends ServerPlugin {
     ResolvedUnitResult analysisResult,
   ) {
     try {
-      if (analysisResult.unit != null) {
+      final path = analysisResult.path;
+      if (analysisResult.unit != null &&
+          path != null &&
+          (driver.analysisContext?.contextRoot.isAnalyzed(path) ?? false)) {
```
incendial, line 180: Do we really need to check this? Seems that `analysisResult.unit` covers this case.

lib/src/config_builder/config_builder.dart
```diff
@@ -12,8 +12,6 @@ import 'models/config.dart';
 
 class ConfigBuilder {
   static const _defaultSkippedFolders = [
-    'ios/',
-    '.dart_tool/**',
     'packages/**',
```
incendial, line 15: So what about packages? Why it's not .packages file?

lib/src/analyzer_plugin/analyzer_plugin.dart
incendial: And why not to check it for cli?

incendial (changes requested):

incendial (approved):

Issue #246, opened 2021-04-03
[Question] What is the proper usage of the GitHub reporter?


GroovinChip:
[This page](https://github.com/dart-code-checker/dart-code-metrics/blob/master/doc/reporters/github-reporter.md) doesn't make it very clear how one might use the GitHub reporter in an Action. What is the proper usage?

incendial:
@dkrutskikh could you please help with that?

@GroovinChip AFAIK, we have it enabled and that is all config that I see right now https://github.com/dart-code-checker/dart-code-metrics/blob/master/.github/workflows/build.yaml#L33.

dkrutskikh:
@GroovinChip after finish work on null-safety stable release. We start works on Github action. At this time it's simple report whose report only about issues.

GroovinChip:
That's fine, I'm just unsure of how to use it at all. When I select it as the reporter using the CLI, the output doesn't seem to be recorded anywhere but the console. And the docs weren't clear either.

Looking forward to actions improvements after null safety is released :) thanks!

GroovinChip:
@dkrutskikh Since null safety has been released to stable, do you have any thoughts on when code metrics for actions will see some progress?

furaiev:
Hello, @dkrutskikh @incendial ,
Thank you for this tool!

I have an issue with git report.
When I run
```
- name: Run Code Metrics
   run: dart run bin/metrics.dart --reporter=github lib
```
I get an error that bin/metrics.dart isn't found.

incendial:
@GroovinChip we merged the PR with the updated docs, please check it out.

Great thanks to @furaiev for creating the PR! 🎉

PR #273, opened 2021-04-21
Added GitHub reporter steps

furaiev:

[ x ] Documentation update

closes https://github.com/dart-code-checker/dart-code-metrics/issues/246

GitHub reporter documentation

### DIFF
--- a/doc/reporters/github-reporter.md
+++ b/doc/reporters/github-reporter.md
@@ -2,18 +2,74 @@
 
 Report about design and static code diagnostics issues in pull requests based on GitHub Actions Workflow commands.
 
-1. Your git workflow must include/support flutter/dart
-2. Add to your git workflow:
-```
+1. Provide dart/flutter and get packages:
+
+**Flutter example**
+
+```yaml
 jobs:
   your_job_name:
-    ...
     steps:
-     ...
+      - name: Install Flutter
+        uses: subosito/flutter-action@master
+        with:
+          channel: stable
+          
+      - name: Install dependencies
+        run: flutter pub get
+      ...
+```
+
+**Dart example**
+
+```yaml
+jobs:
+  your_job_name:
+    steps:
+      - name: Install Dart
+        uses: dart-lang/setup-dart@v1
+      
+      - name: Install dependencies
+        run: flutter pub get
+      ...
+```
+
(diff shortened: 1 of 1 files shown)


furaiev:
Just found that it works only if I copy in the root of my project `bin/metrics.dart` from the package. 
Please advise how to adjust the instructions.

doc/reporters/github-reporter.md
```diff
@@ -2,6 +2,20 @@
 
 Report about design and static code diagnostics issues in pull requests based on GitHub Actions Workflow commands.
 
+1. Your git workflow must include/support flutter/dart
+2. Add to your git workflow:
+```
```
dkrutskikh, line 7: @furaiev please add empty line before backticks, and add language specification for quoted block

dkrutskikh (changes requested):

doc/reporters/github-reporter.md
```diff
@@ -2,6 +2,20 @@
 
 Report about design and static code diagnostics issues in pull requests based on GitHub Actions Workflow commands.
 
+1. Your git workflow must include/support flutter/dart
```
incendial, line 5: It might be not so clear for the external user what does this phrase mean. Seems like they might expect an example here.

doc/reporters/github-reporter.md
```diff
@@ -2,6 +2,20 @@
...
+2. Add to your git workflow:
+```
+jobs:
+  your_job_name:
+    ...
+    steps:
+     ...
+      - name: Run Code Metrics
+        run: flutter pub run dart_code_metrics:metrics lib --reporter=github lib
```
incendial, line 14: This works if the project is added to `dev_dependencies` , but it seems that we should also cover a case when the project is installed as a global dependency.

incendial (changes requested):
First of all, thank you for your contribution. We definitely have some problems with docs and your input is very helpful. If you have some time, I'd suggest to make small improvements to help users even more, but it's okay if you don't have it.

If you don't have time to fix it, just let us know and we will fix the PR by ourselves.

dkrutskikh (approved):

incendial (approved):

PR #287, opened 2021-04-26
refactor: Migrate from InternalResolvedUnitResult

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[X] Other, please explain: migrate from internal InternalResolvedUnitResult


incendial:
@dkrutskikh  I generally thinking that we should keep our own `Unit` model which will have not null unit and content models. What do you think?

PR #291, opened 2021-04-27
Deprecate lines of executable code

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[X] Other, please explain: deprecate lines-of-executable-code


lib/src/obsoleted/analyzer_plugin/analyzer_plugin_utils.dart
```diff
@@ -106,6 +109,33 @@ plugin.AnalysisErrorFixes metricReportToAnalysisErrorFixes(
...
+    ...config.antiPatterns.map((pattern) => pattern.id),
+    ...config.metricsConfig.keys,
+  };
+
+  return deprecatedOptions
+      .where((option) => ids.contains(option.deprecated))
+      .map((option) => plugin.AnalysisErrorFixes(plugin.AnalysisError(
+            plugin.AnalysisErrorSeverity.WARNING,
+            plugin.AnalysisErrorType.HINT,
+            plugin.Location(analysisOptionPath, 0, 1, 0, 0, 0, 0),
```
incendial, line 129: It seems that it should extend current message with deprecation warning instead of creating a new one on top of a file.

incendial (changes requested):

lib/src/obsoleted/analyzer_plugin/analyzer_plugin_utils.dart
dkrutskikh: fixed

incendial (approved):

PR #294, opened 2021-04-28
build: setup flutter environment on CI/CD

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[x] Other, please explain: setup Flutter environment on ci/cd

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -17,7 +17,9 @@ jobs:
         shell: bash
 
     steps:
-      - uses: dart-lang/setup-dart@v1
+      - uses: subosito/flutter-action@v1
+        with:
+          channel: 'stable'
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: dart pub get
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -49,9 +51,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: dart-lang/setup-dart@v1
+      - uses: subosito/flutter-action@v1
         with:
-          sdk: 2.12.0
+          flutter-version: '2.0.0'
```
incendial, line 56: @dkrutskikh are you sure that the version should pinned to a lowest one? Maybe there is a way to always test it with the latest version?

incendial (changes requested):

incendial (approved):

PR #300, opened 2021-04-30
[WIP] fix: support monorepo-like repositories with single root "analysis_options.yaml" file

roman-petrov:
**This is a WIP. I also plan to provide complete test application example here.**

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[x] Add something to the core
[ ] Other, please explain:

Some projects (for example [fluttercommunity/plus_plugins](https://github.com/fluttercommunity/plus_plugins)) have monorepo like layout, for example:

```
packages
- package-1
- package-2
...
- package-N
analysis_options.yaml
```

Currently `dart-code-metrics` incorrectly resolves exclude paths in this scenario: it uses `package` root directory instead of using root workspace/project directory but dart analyzer resolves these paths correctly.

I changed exclude path resolution logic: in the scenario I described above exclude paths are resolved relative to `analysis_options.yaml` directory.

I am not sure if my change can break something. As there is currently no good test coverage for code I modified (and it's hard to get good coverage here due to integration bounds) the only way I see to prevent regressions is code review and manual testing.

### DIFF
--- a/lib/src/analyzer_plugin/analyzer_plugin.dart
+++ b/lib/src/analyzer_plugin/analyzer_plugin.dart
@@ -16,6 +16,7 @@ import 'package:analyzer/src/dart/analysis/file_state.dart';
 import 'package:analyzer_plugin/plugin/plugin.dart';
 import 'package:analyzer_plugin/protocol/protocol_generated.dart' as plugin;
 import 'package:source_span/source_span.dart';
+import 'package:path/path.dart' as path;
 
 import '../analyzers/lint_analyzer/anti_patterns/anti_patterns_factory.dart';
 import '../analyzers/lint_analyzer/metrics/metric_utils.dart';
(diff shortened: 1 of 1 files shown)


lib/src/obsoleted/analyzer_plugin/analyzer_plugin.dart
```diff
@@ -100,20 +101,27 @@ class MetricsAnalyzerPlugin extends ServerPlugin {
       return dartDriver;
     }
 
+    final analysisOptionsFilePath = contextRoot.optionsFile;
```
roman-petrov, line 104: Should we put all this logic into a separate (pure) function and add tests?

lib/src/obsoleted/analyzer_plugin/analyzer_plugin.dart
```diff
@@ -100,20 +101,27 @@ class MetricsAnalyzerPlugin extends ServerPlugin {
...
+    // We need to take into account mono repositories / multiroot workspaces where
+    // "rootPath" is not the same as "analysisOptions.yaml" file location.
+    final excludeRootPath = analysisOptionsFilePath == null
```
roman-petrov, line 107: `contextRoot,optionsFile` is nullable field but I do not know scenarios when it can be null. In this case `rootPath` is still used as root for excludes but I am not sure if this is correct.

roman-petrov:
In most repositories files are ignored by glob pattern, not file path. So it would be hard to test this change in real world project and I decided to close this PR for now.

PR #302, opened 2021-05-01
Update internal unit result usage

incendial:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[X] Other, please explain: refactoring


lib/src/obsoleted/metrics_analyzer.dart
```diff
@@ -269,15 +283,9 @@ class MetricsAnalyzer {
 
         final ignores = Suppression(result.content ?? '', lineInfo);
 
-        final source = InternalResolvedUnitResult(
-          Uri.parse(filePath),
```
incendial, line 273: @dkrutskikh why you didn't take `uri` from `result.uri`?

lib/src/metrics/source_lines_of_code/source_lines_of_code_metric.dart
```diff
@@ -73,11 +73,11 @@ class SourceLinesOfCodeMetric extends FunctionMetric<int> {
   Iterable<ContextMessage> _context(
     Declaration node,
     Iterable<int> linesWithCode,
-    ResolvedUnitResult source,
+    InternalResolvedUnitResult source,
   ) =>
       linesWithCode.map((lineIndex) {
         final lineStartLocation = SourceLocation(
-          source.unit?.lineInfo?.getOffsetOfLine(lineIndex) ?? 0,
+          source.unit.lineInfo?.getOffsetOfLine(lineIndex) ?? 0,
```
dkrutskikh, line 80: @incendial I think we can take `lineInfo` directly from `source`, need to check all code base to find similar situations.

dkrutskikh (changes requested):

lib/src/obsoleted/rules/prefer_trailing_comma.dart
```diff
@@ -29,10 +29,10 @@ class PreferTrailingComma extends ObsoleteRule {
         );
 
   @override
-  Iterable<Issue> check(ResolvedUnitResult source) {
-    final visitor = _Visitor(source.unit!.lineInfo!, _itemsBreakpoint);
+  Iterable<Issue> check(InternalResolvedUnitResult source) {
+    final visitor = _Visitor(source.unit.lineInfo!, _itemsBreakpoint);
```
dkrutskikh, line 33: @incendial use `source.lineInfo` instead `source.unit.lineInfo!`

dkrutskikh (changes requested):

lib/src/metrics/source_lines_of_code/source_lines_of_code_metric.dart
incendial: Done, I've also added checks for `ResultState` to prevent possible exceptions when `lineInfo` is accessed directly.

lib/src/obsoleted/rules/prefer_trailing_comma.dart
incendial: Done.

dkrutskikh (approved):

PR #324, opened 2021-05-10
refactor: create separate config builder

incendial:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[X] Other, please explain: refactor config

### DIFF
--- a/example/example.dart
+++ b/example/example.dart
@@ -1,9 +1,9 @@
 // ignore_for_file: deprecated_member_use_from_same_package
 import 'dart:io';
 
+import 'package:dart_code_metrics/config.dart';
 import 'package:dart_code_metrics/metrics_analyzer.dart';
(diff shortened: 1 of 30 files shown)


example/example.dart
```diff
@@ -3,6 +3,7 @@ import 'dart:io';
 
 import 'package:dart_code_metrics/metrics_analyzer.dart';
 import 'package:dart_code_metrics/reporters.dart';
+import 'package:dart_code_metrics/src/config_builder/config_builder.dart';
```
dkrutskikh, line 6: @incendial import from src, not a good idea for example

dkrutskikh (changes requested):

dkrutskikh (approved):

PR #330, opened 2021-05-12
feat: add `ignored-annotations` option to `avoid-returning-widgets` rule to allow using libraries like "functional_widget"

roman-petrov:
We use [functional_widget](https://pub.dev/packages/functional_widget) library in our Flutter projects to avoid boilerplate code when creating simple stateless flutter widgets. 

In such case there is nothing bad in returning widgets from function because stateless widget class is being generated. So I suggest to add `ignored-annotations` option to `avoid-returning-widgets` rule to allow to use `functional_widget` library without having to always ignore rule violation.

### DIFF
--- a/doc/rules/avoid-returning-widgets.md
+++ b/doc/rules/avoid-returning-widgets.md
@@ -33,9 +33,11 @@ dart_code_metrics:
     - avoid-returning-widgets:
         ignored-names:
           - testFunction
+        ignored-annotations:
+          - FunctionalWidget
 ```
 
-will ignore all functions named `testFunction`.
+will ignore all functions named `testFunction` and all functions having `FunctionalWidget` annotation.
 
 ### Example
 
--- a/doc/rules/avoid-returning-widgets.md
+++ b/doc/rules/avoid-returning-widgets.md
@@ -8,7 +8,10 @@ avoid-returning-widgets
 
 ## Description
 
-Warns when a method or function returns a Widget or subclass of a Widget (**Note** `build` method will not trigger the rule).
+Warns when a method or function returns a Widget or subclass of a Widget. The following patterns will not trigger the rule:
+
+- Widget `build` method overrides.
+- Functions with [functional_widget](https://pub.dev/packages/functional_widget) package annotations.
 
 Extracting widgets to a method is considered as a Flutter anti-pattern, because when Flutter rebuilds widget tree, it calls the function all the time, making more processor time for the operations.
 
@@ -21,7 +24,7 @@ Additional resources:
 * <https://www.reddit.com/r/FlutterDev/comments/avhvco/extracting_widgets_to_a_function_is_not_an/>
 * <https://medium.com/flutter-community/splitting-widgets-to-methods-is-a-performance-antipattern-16aa3fb4026c>
 
-Use `ignored-names` configuration, if you want to ignore a function or method name. For example:
+Use `ignored-names` configuration, if you want to ignore a function or method name. Use `ignored-annotations` configuration, if you want to override default ignored annotation list. For example:
 
 ### Config example
 
@@ -34,10 +37,10 @@ dart_code_metrics:
         ignored-names:
           - testFunction
         ignored-annotations:
-          - FunctionalWidget
+          - allowedAnnotation
 ```
 
-will ignore all functions named `testFunction` and all functions having `FunctionalWidget` annotation.
+will ignore all functions named `testFunction` and all functions having `allowedAnnotation` annotation.
 
 ### Example
 
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
@@ -2,10 +2,19 @@ part of 'avoid_returning_widgets.dart';
 
 class _ConfigParser {
   static const _ignoredNamesConfig = 'ignored-names';
+  static const _ignoredAnnotationsConfig = 'ignored-annotations';
 
   static Iterable<String> getIgnoredNames(Map<String, Object> config) =>
-      config.containsKey(_ignoredNamesConfig) &&
-              config[_ignoredNamesConfig] is Iterable
-          ? List<String>.from(config[_ignoredNamesConfig] as Iterable)
+      _getIterable(config, _ignoredNamesConfig);
+
+  static Iterable<String> getIgnoredAnnotations(Map<String, Object> config) =>
+      _getIterable(config, _ignoredAnnotationsConfig);
+
+  static Iterable<String> _getIterable(
+    Map<String, Object> config,
+    String name,
+  ) =>
+      config.containsKey(name) && config[name] is Iterable
+          ? List<String>.from(config[name] as Iterable)
           : <String>[];
 }
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
@@ -4,17 +4,24 @@ class _ConfigParser {
   static const _ignoredNamesConfig = 'ignored-names';
   static const _ignoredAnnotationsConfig = 'ignored-annotations';
 
+  static const _defaultIgnoredAnnotations = [
+    'FunctionalWidget',
+    'swidget',
+    'hwidget',
+  ];
+
   static Iterable<String> getIgnoredNames(Map<String, Object> config) =>
-      _getIterable(config, _ignoredNamesConfig);
+      _getIterable(config, _ignoredNamesConfig) ?? [];
 
   static Iterable<String> getIgnoredAnnotations(Map<String, Object> config) =>
-      _getIterable(config, _ignoredAnnotationsConfig);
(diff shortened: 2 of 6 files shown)


lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/visitor.dart
```diff
@@ -4,29 +4,51 @@ class _Visitor extends RecursiveAstVisitor<void> {
...
   Iterable<Declaration> get declarations => _declarations;
 
-  _Visitor(this._ignoredNames);
+  _Visitor(this._ignoredNames, this._ignoredAnnotations);
 
   @override
   void visitMethodDeclaration(MethodDeclaration node) {
     super.visitMethodDeclaration(node);
 
-    if (!node.isSetter && !_isIgnored(node.name.name)) {
```
roman-petrov, line 16: I found two duplicate code blocks in `visitMethodDeclaration` and `visitFunctionDeclaration` so decided to move common implementation into `_visitDeclaration` method.

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
```diff
@@ -2,10 +2,19 @@ part of 'avoid_returning_widgets.dart';
...
   static Iterable<String> getIgnoredNames(Map<String, Object> config) =>
-      config.containsKey(_ignoredNamesConfig) &&
-              config[_ignoredNamesConfig] is Iterable
-          ? List<String>.from(config[_ignoredNamesConfig] as Iterable)
+      _getIterable(config, _ignoredNamesConfig);
+
+  static Iterable<String> getIgnoredAnnotations(Map<String, Object> config) =>
+      _getIterable(config, _ignoredAnnotationsConfig);
+
+  static Iterable<String> _getIterable(
```
roman-petrov, line 13: Maybe, it would be great to have common config parsing code somewhere.... Not sure about architecture since I am just starting contributing to the project.

test/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/examples/example.dart
```diff
@@ -42,6 +42,20 @@ Widget _getWidget() => Container();
 
 String _getString() => '';
 
+// LINT
```
roman-petrov, line 45: I added two `// LINT` comments (as I see above in this file) but I don't understand the purpose of this, can you please explain?... Maybe, we should add some information about creating / modifying rule tests in contribution guide?

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
incendial: It sure will, but I think it's not that important right now. If you want, you can add it to `rule_utils.dart` as this functionality is used inside rules for now and updated all `_ConfigParser`'s.

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/visitor.dart
incendial: Yeah, looks good!

test/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/examples/example.dart
incendial: From my point of view the purpose of a `// LINT` comment is to allow you to understand which lines will trigger the rule without seeing the test code or rule implementation. If it's not helping, then we really need to rethink this approach.

test/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/avoid_returning_widgets_test.dart
```diff
@@ -64,6 +70,10 @@ void main() {
           '_getWidgetFuture',
           '_getWidget',
         ],
+        'ignored-annotations': [
+          'FunctionalWidget',
+          'swidget',
+        ]
```
incendial, line 76: Please add a trailing comma here.

incendial (changes requested):
LGTM, although a small change in the docs is required. What do you think, maybe we need to add `FunctionalWidget` annotation to a list by default or it's an unexpected behaviour?

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
roman-petrov: Ok, lets leave it as is for now.

test/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/examples/example.dart
roman-petrov: I thought there is some automatic mechanism for checking these comments... Now I got it: these comments have only informational purpose. I think it still good to have them but I see a couple drawbacks:
- there is no any automation to control if these comments synchronized with actual rule implementation (and new contributors can just easily forget to add such comments)
- these comments are rule configuration specific so generally they apply only to rule with default configuration

Anyway, at the moment I also do not see how to improve this, so we might rethink this in future.

roman-petrov:
> LGTM, although a small change in the docs is required. What do you think, maybe we need to add `FunctionalWidget` annotation to a list by default or it's an unexpected behaviour?

hmm.. I think it is a pretty good idea to allow functional widget annotations in default list: I actually do not see any negative impact on this. I will add all functional widget annotations (`FunctionalWidget`, `swidget` and `hwidget`) to default allow list and update rule documentation.

roman-petrov:
@incendial , I am thinking about your proposal and it looks like in this PR I am adding unnecessary complexity for the rule by introducing additional configuration parameter. At the moment we do not have any other cases with annotations where rule should be ignored.  So  I propose to:
- remove this new configuration parameter
- just hard-code these allowed functional widget decorations in some constant array in code
- update documentation and say, that the rule works correctly with `functional_widget` package

What do you think?

incendial:
@roman-petrov you right that we don't have them at the moment, but I think that it's more flexible to have an option to provide other annotation as well (ex. someone will create a new lib with annotations and our uses will get false-positives that will require our attention).
So, we should have this option, if it's not passed, it should have the default config.

roman-petrov:
> So, we should have this option, if it's not passed, it should have the default config.

Ok, I will move this way and address review comments.

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
```diff
@@ -2,10 +2,26 @@ part of 'avoid_returning_widgets.dart';
...
+
+  static Iterable<String> getIgnoredAnnotations(Map<String, Object> config) =>
+      _getIterable(config, _ignoredAnnotationsConfig) ??
+      _defaultIgnoredAnnotations;
+
+  static Iterable<String>? _getIterable(
+    Map<String, Object> config,
+    String name,
+  ) =>
+      config[name] is Iterable
```
roman-petrov, line 24: Previously we had `config.containsKey(_ignoredNamesConfig) && config[_ignoredNamesConfig] is Iterable` condition here. I removed `config.containsKey(_ignoredNamesConfig)` validation since it seems to be redundant or am I missing something?

doc/rules/avoid-returning-widgets.md
```diff
@@ -33,9 +33,11 @@ dart_code_metrics:
     - avoid-returning-widgets:
         ignored-names:
           - testFunction
+        ignored-annotations:
+          - FunctionalWidget
 ```
 
-will ignore all functions named `testFunction`.
+will ignore all functions named `testFunction` and all functions having `FunctionalWidget` annotation.
```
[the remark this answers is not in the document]
roman-petrov: done

test/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/avoid_returning_widgets_test.dart
roman-petrov: done

roman-petrov:
@incendial, I addressed your comments, please take another look.

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
incendial: Yeah, seems like that. I will fix all other places some time after this PR will be merged. Could you please add `TODO:` for me?

incendial (changes requested):
@roman-petrov could you add

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/config_parser.dart
roman-petrov: I added a PR https://github.com/dart-code-checker/dart-code-metrics/pull/342 instead of `TODO`, I performed a "`containsKey`" search in the project and it seems I already fixed all similar (only three) cases in this new PR.

incendial (approved):

PR #340, opened 2021-05-18
Fix test file name

incendial:

[ ] Documentation update
[X] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

It seems `resolveFile` not checks whether a file exist or not. I've added this check to FileResolver in order to catch this type of problems.


test/helpers/file_resolver.dart
```diff
@@ -9,7 +9,13 @@ class FileResolver {
...
     String filePath,
   ) async {
-    final path = normalize(File(filePath).absolute.path);
+    final file = File(filePath);
+
+    if (!file.existsSync()) {
+      throw StateError('Unable to find a file for the given path: $filePath');
+    }
+
+    final path = normalize(file.absolute.path);
```
dkrutskikh, line 18: @incendial why do we need normalize absolute path?

incendial: @dkrutskikh we have test cases where normalization is needed. So this solution works for all cases.

dkrutskikh (approved):

Issue #313, opened 2021-05-05
[New feature: blocked] Run `dart_code_metrics` when running `dart analyze` command


roman-petrov:
Actually, this is not just a question and closer to feature request. But there is almost nothing should be done in `dart_code_metrics` to implement this feature. 

A relatively new `dart analyze`  command (which replaces `dartanalyzer`) uses analysis server to perform linting. `dart` tool [launches analysis server](https://github.com/dart-lang/sdk/blob/master/pkg/dartdev/lib/src/commands/analyze.dart#L115) and communicates with it to get code analysis issues. I actually spent some time and investigated this scenario. I see that plugins (and `dart_code_metrics` plugin) in such scenario are being loaded and work properly.

But issues found by `dart_code_metrics` are not reported in this scenario: analysis server shutted down too early by `dart` tool. 

There is an issue in Dart SDK repository which blocks us from using the plugin inside `dart analyze` workflow:
[`dart analyze` does not know when analyzer plugin results are complete](https://github.com/dart-lang/sdk/issues/38407)

I decided to open this issue just to track Dart SDK issue .  The issue has `P3` status and I hope it will be fixed relatively soon. My proposal is document ability to run `dart_code_metrics` using `dart analyze` after Dart SDK issue is fixed.  

I suppose it would be more convenient to use the plugin with  "standard" workflow with `dart analyze` , please let me know what do you think about this.

incendial:
Yeah, we faced situations when running `dart analyze` could eventually fail because of a plugin issue report ([like this one](https://github.com/dart-code-checker/dart-code-metrics/runs/2311793200?check_suite_focus=true#step:5:6)), but unfortunately had no time to investigate it further or even help fix it.

I think it'd be great to have this feature.

incendial:
But we have other ideas for Cli commands, so we will not be able to remove it completely.

roman-petrov:
> But we have other ideas for Cli commands, so we will not be able to remove it completely.

Actually, I am proposing just to have new feature and document it when Dart SDK issue is fixed. I understand that CLI integration can give us more than just analyzer plugin so I am not proposing to remove CLI commands.

Let's have this issue open if you don't mind: I will try to track it myself and hopefully will provide documentation PR when DartSDK issue is resolved.

incendial:
Considering the overall direction with the DCM as a separate executable, this feature request is now obsolete.

PR #344, opened 2021-05-18
[DO_NOT_MERGE] research: CLI performance issue profiling

roman-petrov:
I started integrating the project into our codebase and found that CLI runs extremely slow compared to `dart analyze` command. 

`dart analyze` command takes about 3-5 seconds on our codebase, `dart_code_metrics` CLI runs about two minutes.

I am not sure if this is some environment configuration or maybe Windows issue. On my machine results of this simple profiling code are:

`contextForMs: 76, getResolvedUnitMs: 115290, runAnalysisforFileMs: 128`.

So `analysisContext.currentSession.getResolvedUnit` method call seems to be **extremely** slow (in another performance test I made I see that it might even take several seconds to complete for just one file). At the moment I do not know how to speed up this...

@incendial, @dkrutskikh - can you please check (when you have time) if this issue can be reproduced on your machines and if we actually have serious performance issue here?

### DIFF
--- a/lib/src/analyzers/lint_analyzer/lint_analyzer.dart
+++ b/lib/src/analyzers/lint_analyzer/lint_analyzer.dart
@@ -95,24 +95,37 @@ class LintAnalyzer {
 
     final analyzerResult = <FileReport>[];
 
+    var totalContextForMs = 0;
+    var totalGetResolvedUnitMs = 0;
+    var totalRunAnalysisforFileMs = 0;
+    DateTime startTime;
     for (final filePath in filePaths) {
       final normalized = normalize(absolute(filePath));
 
+      startTime = DateTime.now();
       final analysisContext = collection.contextFor(normalized);
+      totalContextForMs += DateTime.now().difference(startTime).inMilliseconds;
+
+      startTime = DateTime.now();
       final unit =
           // ignore: deprecated_member_use
           await analysisContext.currentSession.getResolvedUnit(normalized);
+      totalGetResolvedUnitMs += DateTime.now().difference(startTime).inMilliseconds;
+
+      startTime = DateTime.now();
       final result = _runAnalysisForFile(
         unit,
         config,
         rootFolder,
         filePath: filePath,
       );
+      totalRunAnalysisforFileMs += DateTime.now().difference(startTime).inMilliseconds;
 
       if (result != null) {
         analyzerResult.add(result);
       }
     }
+    print('contextForMs: $totalContextForMs, getResolvedUnitMs: $totalGetResolvedUnitMs, runAnalysisforFileMs: $totalRunAnalysisforFileMs');
 
     return analyzerResult;
   }


dkrutskikh:
@roman-petrov please try to migrate on getResolvedUnit2

roman-petrov:
@dkrutskikh, I already tried to migrate to `getResolvedUnit2` and saw no any difference in terms of performance. I think we should check if this performance issue exists and can be easily reproduced first and then decide if it possible to fix it...

incendial:
@roman-petrov the problem is that this is the only _known_ way to get info about types. We can use `currentSession.getParsedUnit2` instead and it will improve the performance a lot (the result of running tests is ~5 times faster `00:43 +205 -6: Some tests failed. `). But for those 6 cases we need a type info.

I will take a closer look later.

roman-petrov:
@incendial, thank you for the clarification. I understand that this not an easy to solve issue. But currently due to this issue CLI speed might be not acceptable in relatively large code base...

I see several ways to improve this:
- Use `getParsedUnit2` for cases/rules we don't need type info. 
- Investigate how analysis works in Dart analyzer (through server): it has good performance, but I am not sure it uses type information
- Cooperate with DartSDK team to find out how we can improve analysis performance (maybe there is some different API we should use for our scenario)
 
It seems `getResolvedUnit2` in our scenario works a bit strange at the moment and it seems Dart compiler does too much work not taking advantage of caching. 

I will also continue investigating ways to solve this issue and will keep this PR open if you don't mind (or we can convert it to issue).

incendial:
Also we can investigate https://github.com/dart-code-checker/dart-code-metrics/issues/313 and it can solve the problem. Either way it requires a deeper research.

roman-petrov:
sure, #313 may partially solve the problem (after the underlying Dart SDK issue is fixed) but we still have it when running in dart code metrics CLI

incendial:
> but we still have it when running in dart code metrics CLI

Agreed. I'll take a closer look after 4.0 will be released.

roman-petrov:
@incendial - at the momentI do not have time and ideas that would quickly fix this performance problem. Maybe it will be better to close this PR for now?

roman-petrov:
ok, let's leave it here for now. I also may put some effort into this in the future since linting speed is important for large / relatively large codebases.

Although I have a quick idea. You have a [GitHub acition](https://github.com/dart-code-checker/dart-code-metrics-action) but it is not integrated into the dart code metrics repository itself. Maybe, it would be better to add this action first to "see" CLI performance.

incendial:
@roman-petrov is it a blocker for you to start using the package?

roman-petrov:
@incendial, yes, the performance issue is almost blocking us from starting. But at the moment it is not critical for us: we have a huge amount of other things to do while migrating our UI development technology stack to Flutter.

incendial:
@roman-petrov then I need to know what is a "good enough" performance for you and are you ready to not being able to use some analysis lints if it will be the only way to fix the problem?

incendial:
I mean, some rules just won't work without full analysis (especially, flutter rules).

roman-petrov:
> @roman-petrov then I need to know what is a "good enough" performance for you and are you ready to not being able to use some analysis lints if it will be the only way to fix the problem?

Actually, I don't have an exact answer to this question. I would say something like: the performance will be good enough if the time `dart_code_metrics` takes while analyzing the project will have the same order that consumes `dart analyze` will almost all lint rules enabled.

For example, if `dart analyze` takes 2 seconds, then it would be ok if `dart_code_metrics` takes something like one, two, or even five seconds. But 20 seconds, two minutes, five minutes might be not acceptable in terms of effective workflow.

roman-petrov:
Actually, as I understand, the performance issue we discuss is somehow related to `getResolvedUnit` method in `analyzer` package. So we don't have to optimize rules themselves, we should think about optimizing program flow somehow.

Just another "draft" idea: `dart analyze` even when running from CLI uses analysis server. Maybe the same can be done in `dart_code_metrics` CLI... I am not familiar enough with the dart analyzer, so I am not sure if this approach can help.

incendial:
> For example, if dart analyze takes 2 seconds, then it would be ok if dart_code_metrics takes something like one, two, or even five seconds. But 20 seconds, two minutes, five minutes might be not acceptable in terms of effective workflow.

Got it, sounds pretty reasonable.

> Just another "draft" idea: dart analyze even when running from CLI uses analysis server. Maybe the same can be done in dart_code_metrics CLI... I am not familiar enough with the dart analyzer, so I am not sure if this approach can help.

Yeah, thats one way to look at it, but we definitely need a more deep research.

incendial:
I think we finally find out what was the problem, closing due to: https://github.com/dart-code-checker/dart-code-metrics/pull/428.

PR #347, opened 2021-05-19
Support all variables for no magic number instead of only constants

incendial:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[X] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:


dkrutskikh:
@incendial please port this changes to master branch.

PR #373, opened 2021-06-04
Upport

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:


doc/rules/no-magic-number.md
```diff
@@ -8,9 +8,7 @@ no-magic-number
 
 ## Description
 
-Warns against using number literals outside of named constants.
-Exception is made for common constants by default: -1, 0 and 1.
-Exception is made for literals inside `DateTime` constructor as there is no way to create `const` `DateTime` and extracting each `int` argument to separate named constant is far too inconvenient
+Warns against using number literals outside of named constants or variables. Exception is made for common constants by default: -1, 0 and 1. Exception is made for literals inside `DateTime` constructor as there is no way to create `const` `DateTime` and extracting each `int` argument to separate named constant is far too inconvenient
```
incendial, line 11: Warns against using number literals outside of named constants or variables. Exceptions are made for common constants (by default: -1, 0 and 1) and for literals inside `DateTime` constructor as there is no way to create `const` `DateTime` and extracting each `int` argument to separate named constant is far too inconvenient.

incendial (changes requested):

doc/rules/no-magic-number.md
dkrutskikh: updated

incendial (approved):

PR #383, opened 2021-06-14
feat: improve long method

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/analyzers/lint_analyzer/anti_patterns/anti_patterns_list/long_method.dart
+++ b/lib/src/analyzers/lint_analyzer/anti_patterns/anti_patterns_list/long_method.dart
@@ -1,12 +1,11 @@
-import 'package:analyzer/dart/ast/ast.dart' as analyzer;
-import 'package:analyzer/dart/element/type.dart' as analyzer;
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/element/type.dart';
 
 import '../../../../utils/node_utils.dart';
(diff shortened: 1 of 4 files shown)


lib/src/analyzers/lint_analyzer/anti_patterns/anti_patterns_list/long_method.dart
```diff
@@ -1,3 +1,6 @@
+import 'package:analyzer/dart/ast/ast.dart' as analyzer;
+import 'package:analyzer/dart/element/type.dart' as analyzer;
```
incendial, line 2: Let's not use named imports here since we don't use them anywhere else.

lib/src/analyzers/lint_analyzer/anti_patterns/anti_patterns_list/long_method.dart
```diff
@@ -37,6 +41,10 @@ class LongMethod extends ObsoletePattern {
     final issues = <Issue>[];
 
     for (final function in functions) {
```
incendial, line 43: I'd suggest to use where instead: `for (final function in functions.where(_isExcluded).toList())`

incendial (changes requested):

incendial (approved):

PR #394, opened 2021-06-23
fix: correctly handle generics for always remove listeners rule

incendial:

[ ] Documentation update
[X] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:


lib/src/analyzers/lint_analyzer/rules/rules_list/always_remove_listener/visitor.dart
```diff
@@ -154,7 +154,9 @@ class _Visitor extends RecursiveAstVisitor<void> {
 
     return removedTarget is Identifier &&
         removedTarget.name == targetName &&
-        removedTarget.staticElement == staticElement;
+        (removedTarget.staticElement == staticElement ||
```
dkrutskikh, line 157: @incendial what if `staticElement` be null?

dkrutskikh (approved):

PR #405, opened 2021-07-18
fix: fix unused files analyzer files extraction

incendial:

[ ] Documentation update
[X] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/analyzers/unused_files_analyzer/unused_files_analyzer.dart
+++ b/lib/src/analyzers/unused_files_analyzer/unused_files_analyzer.dart
@@ -50,8 +50,13 @@ class UnusedFilesAnalyzer {
       final unusedFilesAnalysisConfig =
           ConfigBuilder.getUnusedFilesConfig(contextConfig, rootFolder);
 
+      final contextFolders = folders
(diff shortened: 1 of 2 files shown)


lib/src/analyzers/unused_files_analyzer/unused_files_analyzer.dart
```diff
@@ -50,8 +50,13 @@ class UnusedFilesAnalyzer {
       final unusedFilesAnalysisConfig =
           ConfigBuilder.getUnusedFilesConfig(contextConfig, rootFolder);
 
+      final contextFolders = folders
+          .where((path) => normalize(join(rootFolder, path))
```
dkrutskikh, line 54: @incendial looks similar than code at 37 line. And what about similar bloc at lib/src/analyzers/lint_analyzer/lint_analyzer.dart ?

dkrutskikh (changes requested):

PR #430, opened 2021-08-20
[WIP] feat: implement `--sdk-path` CLI option to use when running as Window…

### LINKED ISSUE #385: [BUG] Compiling CLI to exe on Windows
Today we discussed CLI analysis performance with @dkrutskikh and decided to try to compile CLI tool to `exe` on Windows (using `dart compile exe`) as a possible workaround.

Compilation works fine, but when I try to launch analysis with created executable like `metrics.exe C:\Projects\dart-code-metrics` i get runtime exceptions, here is the output:

```
[exception: FileSystemException(path=C:\Projects\lib\_internal\sdk_library_metadata\lib\libraries.dart; message=Cannot open file)][stackTrace: #0      _PhysicalFile.readAsStringSync (package:analyzer/file_system/physical_file_system.dart:184)
#1      FolderBasedDartSdk.initialLibraryMap (package:analyzer/src/dart/sdk/sdk.dart:473)
#2      new FolderBasedDartSdk (package:analyzer/src/dart/sdk/sdk.dart:371)
#3      ContextBuilder.findSdk.<anonymous closure> (package:analyzer/src/context/builder.dart:198)
#4      _HashMap.putIfAbsent (dart:collection-patch/collection_patch.dart:140)
#5      DartSdkManager.getSdk (package:analyzer/src/generated/sdk.dart:95)
#6      ContextBuilder.findSdk (package:analyzer/src/context/builder.dart:197)
#7      ContextBuilder.createSourceFactoryFromWorkspace (package:analyzer/src/context/builder.dart:167)
#8      ContextBuilder.buildDriver (package:analyzer/src/context/builder.dart:110)
#9      ContextBuilderImpl.createContext (package:analyzer/src/dart/analysis/context_builder.dart:91)
#10     new AnalysisContextCollectionImpl (package:analyzer/src/dart/analysis/analysis_context_collection.dart:63)
#11     LintAnalyzer.runCliAnalysis (package:dart_code_metrics/src/analyzers/lint_analyzer/lint_analyzer.dart:75)
#12     CliRunner.runAnalysis (package:dart_code_metrics/src/cli/cli_runner.dart:21)
<asynchronous suspension>
#13     main (file:///c:/projects/dart-code-metrics/bin/metrics.dart:4)
<asynchronous suspension>
]
[exception: FileSystemException(path=C:\Projects\lib\_internal\libraries.dart; message=Cannot open file)][stackTrace: #0      _PhysicalFile.readAsStringSync (package:analyzer/file_system/physical_file_system.dart:184)
#1      FolderBasedDartSdk.initialLibraryMap (package:analyzer/src/dart/sdk/sdk.dart:473)
#2      new FolderBasedDartSdk (package:analyzer/src/dart/sdk/sdk.dart:371)
#3      ContextBuilder.findSdk.<anonymous closure> (package:analyzer/src/context/builder.dart:198)
#4      _HashMap.putIfAbsent (dart:collection-patch/collection_patch.dart:140)
#5      DartSdkManager.getSdk (package:analyzer/src/generated/sdk.dart:95)
#6      ContextBuilder.findSdk (package:analyzer/src/context/builder.dart:197)
#7      ContextBuilder.createSourceFactoryFromWorkspace (package:analyzer/src/context/builder.dart:167)
#8      ContextBuilder.buildDriver (package:analyzer/src/context/builder.dart:110)
#9      ContextBuilderImpl.createContext (package:analyzer/src/dart/analysis/context_builder.dart:91)
#10     new AnalysisContextCollectionImpl (package:analyzer/src/dart/analysis/analysis_context_collection.dart:63)
#11     LintAnalyzer.runCliAnalysis (package:dart_code_metrics/src/analyzers/lint_analyzer/lint_analyzer.dart:75)
#12     CliRunner.runAnalysis (package:dart_code_metrics/src/cli/cli_runner.dart:21)
<asynchronous suspension>
#13     main (file:///c:/projects/dart-code-metrics/bin/metrics.dart:4)
<asynchronous suspension>
]
```
As you can see, the source of these exceptions is not `dart_code_metrics` itself, they are coming from `analyzer` package. 

Not sure if anything can be done here from our side, it seems we will need changes from Dart SDK team.

I created this issue because @dkrutskikh suggested me to do so, please feel free to just close it if you like :)

roman-petrov:
Closes #385


roman-petrov:
The implementation seems to be working just fine: now I am able to run dart code metrics as windows executable. I did some measurements again to check if compilation to exe improves performance.

I created two batch files:

`measure.bat`:

```cmd
@echo %time%
@call dart dart-code-metrics/bin/metrics.dart analyze dart-code-metrics/lib --sdk-path D:\SDK\dart-sdk >nul
@echo %time%
```

`measure_exe.bat`:
```cmd
@echo %time%
@call metrics.exe analyze --sdk-path D:\SDK\dart-sdk dart-code-metrics/lib >nul
@echo %time%
```

The results I see:

`measure.bat`: ~8 seconds

dart

`measure.exe.bat`: ~ 2 seconds

exe

As you can see, there is an "amazing" performance improvement, but I think this is too good to be true ;) So I think we need more measurements/comparisons to find out if we get real performance benefits when compiling to exe like:

- Do we get performance improvement if we compare exe performance with aot snapshot performance when dart code metrics is activated globally from pub? I was unable to quickly test this because the current version is not published yet, and there is no precompilation when activating from path https://github.com/dart-lang/sdk/issues/46914#issuecomment-901437225
- Can you provide some measurements from your side? @dkrutskikh 
- Maybe it would be good to ask the Telegram community, what they think about this, do they see performance benefit.

So let's have further discussions on this if you don't mind. I will finish this PR if we will decide to move forward with this approach.

incendial:
I think the result is really good and it correlates with what I measured locally (~8 seconds for compilation).

> Do we get performance improvement if we compare exe performance with aot snapshot performance when dart code metrics is activated globally from pub?

Yeah, that's a good question, I think, the answer will not surprise us 😄. We published 4.2.0 today, could you please measure the snapshot version too?

> Maybe it would be good to ask the Telegram community, what they think about this, do they see performance benefit.

Please show me a person who'd say that slower is better 😄😄😄. I think that it's a right direction no matter what.

The only suggestion I have is to check how `dart analyze` handles SDK detection and maybe we can find a solution without a separate option. Could you please research it too?

incendial:
I really like the trend in your tests 12 secs -> 8 secs -> 2 secs. It's such a magnificent improvement!!!

roman-petrov:
@incendial, I like the trend too :) But there is still something we have to do before we get this ready to be used...

Tomorrow I will measure speed of running dart code metrics as precompiled AOT snapshot and publish the result here. We can continue discussing if running as EXE will still be significantly faster, if not - I think we can just stop here.

About Dart SDK path detection: I think, this is actually not possible. Users can have multiple Dart SDK versions installed (like I do). `dart analyze` runs inside Dart process so I assume it is easy to detect Dart SDK path from process handle / `dart.exe` path.

But when we compile to Windows EXE this information is lost and there is no any association between our EXE file and Dart SDK.

So I do not see many options at the moment:
- Use CLI option for SDK path
- Bundle required Dart SDK files as CLI application assets. But even in this case, I am not sure if Dart compiler will bundle these assets into compiled EXE file. 

Actually, second scenario as I understand is not quite easy to implement. So as you see I am not quite optimistic here, but I will be back tomorrow with AOT snapshot measurements :)

roman-petrov:
Today I performed three performance measurements. All measurements were done on this branch. Here are three batch files and the results:

`measure_local.bat`

```cmd
@echo %time%
@call dart dart-code-metrics/bin/metrics.dart analyze dart-code-metrics/lib --sdk-path D:\SDK\dart-sdk >nul
@echo %time%
```

`measure_aot.bat` (this uses globally activated dart code metrics 4.2.0 and I ignored first launch while measuring because there is time to precompile aot)

```cmd
@echo %time%
@call C:\Users\roman.petrov\AppData\Local\Pub\Cache\bin\metrics.bat  analyze dart-code-metrics/lib >nul
@echo %time%
```

`measure_exe.bat`

```cmd
@echo %time%
@call metrics.exe analyze --sdk-path D:\SDK\dart-sdk dart-code-metrics/lib >nul
@echo %time%
```

The results are: 

`measure_local.bat`:  ~8 seconds

local

`measure_aot.bat`: ~4,5 seconds

aot

`measure_exe.bat`: ~2,5 seconds

exe

So exe still seems to be fastest method from my measurements. @dkrutskikh - can you please repeat these measurements on Windows to validate my results?..

If we decide to provide some fast dart code metrics workflow with exe compilation we need to figure out first how this can be done :) Current PR with CLI option might be just a starting point.

roman-petrov:
@incendial,

A couple of days ago @dkrutskikh proposed an idea: we can try to detect SDK path from system PATH environment variable. I think this might be quite a good approach. The logic can be as follows:

* for each <directory> in `PATH` try to find `dart.exe`  
* if found - use parent directory of this <directory> as SDK path

The possible drawback: some users (like I do) can have multiple Dart SDK versions installed and SDK we found might be outdated. In this case, users will just get unpredictable runtime errors when running compiled exe. But I still think that this approach is a good trade-off between complexity and functionality.

roman-petrov:
But the most important question for me is: will Windows users actually use compiled exe? We know, compiling to exe gives a significant performance improvement. But it involves non standard Dart workflow:

* Standard workflow: to use/upgrade dart package it is required only to run `dart pub global activate <package_name>` and that's all.
*  EXE workflow: users will have to clone the repository or activate the package globally first to download the package source code. After that, they should compile it using `dart compile exe`. The upgrade process requires repeating these steps.

So I am completely unsure if users will like the second workflow and use it. This complex workflow might provide major benefits only for large codebases. Personally, I am quite satisfied with CLI performance after https://github.com/dart-code-checker/dart-code-metrics/pull/428.

@incendial , @dkrutskikh  - what do you think, should we continue with this?

incendial:
> The possible drawback: some users (like I do) can have multiple Dart SDK versions installed and SDK we found might be outdated. In this case, users will just get unpredictable runtime errors when running compiled exe. But I still think that this approach is a good trade-off between complexity and functionality.

This exactly why we need to combine those approaches: we will provide auto-detection based on PATH, **but** if it'll not work for the user in a specific case (like having multiple SDK version) we should also provide a cli option in order to let the user get a working exe and not be blocked.

> EXE workflow: users will have to clone the repository or activate the package globally first to download the package source code. After that, they should compile it using dart compile exe. The upgrade process requires repeating these steps.

I think we can choose an approach with package managers and distribute precompiled version like `fvm` does, for example. This approach will not require any action from the user, but to download the precompiled executable and call it.

incendial:
But we can't move to this step until we have no doubts about the package working correctly and without SDK detection it won't work, so I think we should definitely finish this improvement and then consider a distribution through package managers.

roman-petrov:
So: 

- we will use the '--sdk-path` option that I already implemented.
- I will update this PR and implement sdk path detection from `PATH` environment variable: this should work if the application process started on Windows as .`exe` file
- `.exe` distribution will be not part of this PR and should be discussed/implemented in another issue?

@incendial, is that correct?..

incendial:
Yeah, that's how I imagined it. Completing the first two steps will unblock use-cases with local compilation (like you have) and let you speed up the pipeline. Then we'll move forward to improving the DX here with precompiled distribution. 

Will this work?

roman-petrov:
I think this would be ok and I like the result we got. I will continue my work on this PR and implement sdk path detection from `PATH`. 

BTW, I will have a two-week vacation in a few days, so I think this PR will be completed only after that. Please let me know if it's ok or just feel free to take ownership of this branch if you like to deliver it faster.

incendial:
> BTW, I will have a two-week vacation in a few days, so I think this PR will be completed only after that. 

Sure, no problem.

roman-petrov:
@incendial I completed Dart SDK detection functionality here. I tested EXE compilation on my machine and the compiled application seems to be working just fine without `--sdk-path` argument.

Can you please take a look if the detection logic fits our needs?

incendial:
@roman-petrov the docs says that exe can be run on different platforms https://dart.dev/tools/dart-compile#exe and I've tested it on macos and it looks fine. So, If there is nothing else you want to add/change in the current implementation, I think we can merge it.

roman-petrov:
@incendial - it's a good news. Do you have any code comments for current implementation?

I see that `codecov` found coverage regression, I will try to add some tests before we merge.

incendial (commented):
Let's not mention Windows anyhow, I think it should work properly on any platform. Also, could you add the docs link https://dart.dev/tools/dart-compile#exe somewhere?

Everything else is cool, I think we can merge it after you add tests for `validateSdkPath` and `findSdkPath`.

roman-petrov:
@incendial , sure, I will add the link to the `https://dart.dev/tools/dart-compile#exe` somewhere, if I understand correctly, do we need it in user docs?

incendial:
@roman-petrov I think that adding it into the description of the argument should be enough https://github.com/dart-code-checker/dart-code-metrics/pull/430/files#diff-34a9cc97d58062ac8d9ab552da6d66e78e8e3d41a0d1859c7561c8115d5d2366R47

roman-petrov:
@incendial - thank you, I will add the link. In a few days I will update this PR and add tests too.

I still have a question about non-Windows platforms. At the moment as you can see in the [code](https://github.com/dart-code-checker/dart-code-metrics/pull/430/files#diff-34a9cc97d58062ac8d9ab552da6d66e78e8e3d41a0d1859c7561c8115d5d2366R116) Dart SDK detection is explicitly enabled only for Windows platform.

You said you tested this PR on macOS and it works fine too. So I am a bit stuck here, how my implementation can affect macOS... I assume that for non-Windows platforms this PR does nothing more than adding the `--sdk-path` CLI parameter, Dart SDK detection is disabled. Did you perform the test on macOS using this new parameter, can you please clarify?...

incendial:
@roman-petrov okay, now I got your point. Yeah, I've tested compilation **without** the new parameter and the idea "it should work properly on any platform" was about the compilation too. 

So, yeah, let's keep mentioning Windows for now and when we'll return to distributing compiled version we'll update the parameter if needed.

roman-petrov:
@incendial - sorry for late response. I was ill and still recovering, hope to get well and finish this this or next week.

roman-petrov:
I am finally very excited to get well and get back to work:)  I finished working on this, PTAL @incendial , @dkrutskikh 

I addressed review comments, added tests and also moved some duplicated code into `BaseCommand` class. Still see some duplication between three commands we have, but this seems to be out of my pull request scope.

PR #469, opened 2021-09-22
feat: ignore Flutter builder functions in "prefer-extracting-callbacks" rule

roman-petrov:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[x] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

It is a pretty common situation in Flutter to have "builders". Widgets even may have multiple "builder" arguments like `titleBuilder`, `contentBuilder`, etc. We are getting violations from dart code metrics on such arguments so I think it would be ok to ignore such arguments in `prefer-extracting-callbacks` rule.

I will create a separate PR to website repository with documentation updates.  It's a little bit annoying to manually keep this repository documentation and website documentation synchronized, maybe we should start thinking about avoiding this? (we can define primary documentation source somehow, for example)

### DIFF
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_extracting_callbacks/visitor.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_extracting_callbacks/visitor.dart
@@ -57,7 +57,7 @@ class _InstanceCreationVisitor extends RecursiveAstVisitor<void> {
   }
 
   bool _isFlutterBuilder(FunctionExpression expression) {
-    if (!isWidget(expression.declaredElement?.returnType)) {
+    if (!isWidgetOrSubclass(expression.declaredElement?.returnType)) {
       return false;
     }
 
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_extracting_callbacks/visitor.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_extracting_callbacks/visitor.dart
@@ -40,7 +40,8 @@ class _InstanceCreationVisitor extends RecursiveAstVisitor<void> {
 
       if (_isNotIgnored(argument) &&
           expression is FunctionExpression &&
-          _hasNotEmptyBlockBody(expression)) {
+          _hasNotEmptyBlockBody(expression) &&
+          !_isFlutterBuilder(expression)) {
         _expressions.add(argument);
       }
     }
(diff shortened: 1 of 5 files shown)


dkrutskikh:
@roman-petrov I think detect `builder` only by name it's not a good idea. What if we check lambda function arguments and find argument with `BuildContex` type?

roman-petrov:
> @roman-petrov I think detect `builder` only by name it's not a good idea. What if we check lambda function arguments and find argument with `BuildContex` type?

Thank you, I think it might be a good idea to try to identify builders by function argument type. So we can, for example, assume that the argument is a "builder" if its RHS expression is a function with the first argument of type `BuildContext`. So, again, 
the heuristic can be described as **Ignore all arguments which value is a function with the first argument having type `BuildContext`**

In this case we can even avoid checking argument names (we currently ignore arguments named "builder"). So will it be ok if we totally remove the argument name check and ignore "builders" based on the heuristics above, what do you think? 

 @incendial , @dkrutskikh

incendial:
@roman-petrov sorry for the long response from my side, I'll take a closer look on weekends.

incendial:
@roman-petrov I'm not a fan of ignoring several arguments by their name, but the idea about checking for BuildContext looks solid.

roman-petrov:
@incendial , thank you, I will proceed with the new approach.

roman-petrov:
I am done with this, PTAL.

lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_extracting_callbacks/visitor.dart
```diff
@@ -55,8 +56,19 @@ class _InstanceCreationVisitor extends RecursiveAstVisitor<void> {
     return body.block.statements.isNotEmpty;
   }
 
+  bool _isFlutterBuilder(FunctionExpression expression) {
+    if (!isWidget(expression.declaredElement?.returnType)) {
```
dkrutskikh, line 60: @roman-petrov in some cases builder function returns subclass of widget, for example:

https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/cupertino/text_field.dart#L1054 , in this case builder returns Row classю

dkrutskikh (changes requested):

lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_extracting_callbacks/visitor.dart
roman-petrov: Done, please take a look to updated code/test/documentation.

dkrutskikh (approved):

dkrutskikh:
@roman-petrov now all looks good for me, thank you

PR #637, opened 2022-01-10
feat: support excludes for a separate anti-pattern.

dkrutskikh:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/analyzers/lint_analyzer/lint_analyzer.dart
+++ b/lib/src/analyzers/lint_analyzer/lint_analyzer.dart
@@ -187,7 +187,6 @@ class LintAnalyzer {
             ignores,
             internalResult,
             config,
-            filePath,
           ),
         );
       }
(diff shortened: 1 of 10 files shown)


website/docs/getting-started/configuration.md
```diff
@@ -145,7 +145,7 @@ analyzer:
 
 will work both for the analyzer and for this plugin.
 
-If you want a specific rule to ignore files, you can configure `exclude` entry for it. For example,
+If you want a specific rule or anti-pattern to ignore files, you can configure `exclude` entry for it. For example,
```
incendial, line 148: Let's have a separate entry for anti-pattern excludes since it has a different config.

lib/src/analyzers/lint_analyzer/lint_analyzer.dart
```diff
@@ -249,13 +248,12 @@ class LintAnalyzer {
...
     InternalResolvedUnitResult source,
     LintAnalysisConfig config,
-    String filePath,
   ) =>
       config.codeRules
           .where((rule) =>
               !ignores.isSuppressed(rule.id) &&
               !isExcluded(
-                filePath,
+                source.path,
```
incendial, line 256: Are you sure about this removal? Why it is possible?

incendial (changes requested):

incendial (approved):

PR #658, opened 2022-01-20
fix: make `check-unused-l10n` also cover supertype member calls

incendial:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/analyzers/unused_l10n_analyzer/unused_l10n_analyzer.dart
+++ b/lib/src/analyzers/unused_l10n_analyzer/unused_l10n_analyzer.dart
@@ -193,6 +193,8 @@ class UnusedL10nAnalyzer {
         issues: issues,
       );
     }
+
+    return null;
   }
 
   Iterable<UnusedL10nIssue> _getUnusedAccessors(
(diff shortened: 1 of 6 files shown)


lib/src/analyzers/unused_l10n_analyzer/unused_l10n_analyzer.dart
```diff
@@ -128,40 +129,71 @@ class UnusedL10nAnalyzer {
...
+    final filePath = unit.source.toString();
+    final relativePath = relative(filePath, from: rootFolder);
+
+    final issues = [
+      ...accessorSourceSpans,
+      ...methodSourceSpans,
+    ];
+
+    if (issues.isNotEmpty) {
+      reports.add(
```
dkrutskikh, line 186: @incendial can we return object instead add them to passed list? This looks like side effect.

dkrutskikh (changes requested):

lib/src/analyzers/unused_l10n_analyzer/unused_l10n_analyzer.dart
incendial: Fixed

dkrutskikh (approved):

PR #687, opened 2022-02-06
feat: add avoid-dynamic rule

incendial:

[ ] Documentation update
[ ] Bug fix
[X] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/website/docs/rules/common/avoid-dynamic.md
+++ b/website/docs/rules/common/avoid-dynamic.md
@@ -1,0 +1,40 @@
+# Avoid dynamic
+
+## Rule id {#rule-id}
+
+avoid-dynamic
+
+## Severity {#severity}
+
+Warning
+
+## Description {#description}
+
+Warns when `dynamic` type is used as variable type in declaration, return type of a function, etc. Using `dynamic` is considered unsafe since it can easily result in runtime errors.
+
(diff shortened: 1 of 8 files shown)


incendial:
@roman-petrov and thus one too, please.

website/docs/rules/common/avoid-dynamic.md
```diff
@@ -0,0 +1,40 @@
...
+
+avoid-dynamic
+
+## Severity {#severity}
+
+Warning
+
+## Description {#description}
+
+Warns when `dynamic` type is used as variable type in declaration, return type of a function, etc. Using `dynamic` is considered unsafe since it can easily result in runtime errors.
```
roman-petrov, line 13: Actually IMHO using `dynamic` is fine when working with non-typed code. However, I prefer to isolate / wrap such code in my code base: using `dynamic` in implementation and provide only strictly typed public interface / wrapper. 

So, maybe we should add here some "good practice" note here like:

"In some situations you still have to use `dynamic` type and these cases might be pretty valid. In such cases you can just isolate or wrap code which works with dynamic data into a single file or package and disable `avoid-dynamic` rule for this dynamically typed code."

PR #708, opened 2022-02-25
feat: Create new rule `ban-name`

### LINKED ISSUE #702: [New rule] `ban-name` Ban some user-specified functions (Willing to PR)
The user can configure some functions that he wants to ban. For example, in my own app, I want to ban `showDialog` and force to use `myShowDialog` instead. This is because (let me explain the simplified case) I added logging in `myShowDialog`, and I do want every dialog to be logged.

N/A

[x] Warns about a potential error (problem)
[ ] Other (please specify:)

configuration:

```
ban_function:
  showDialog: 'Please use myShowDialog instead'
  ... # other functions to ban
```

```dart
showDialog(...); // LINT WARN!

myShowDialog(...); // good
```

yes

fzyzcjy:

[ ] Documentation update
[ ] Bug fix
[x] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Close #702


lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/ban_name_rule.dart
```diff
@@ -0,0 +1,46 @@
...
+import '../../models/common_rule.dart';
+import '../../rule_utils.dart';
+
+part 'visitor.dart';
+part 'utils/config_parser.dart';
+
+class BanNameRule extends CommonRule {
+  static const String ruleId = 'ban-name';
+
+  final List<_BanNameConfigEntry> _entries;
```
dkrutskikh, line 20: @fzyzcjy please use Set for automatically remove duplicates

dkrutskikh (changes requested):

lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/ban_name_rule.dart
fzyzcjy: Not automatically indeed. _BanNameConfigEntry not only has a name, but also a *description*. So sets will not prevent two entries with same name.

And if that happens, the plugin can still run (and it is indeed user's fault)

test/src/analyzers/lint_analyzer/rules/rules_list/ban_name/ban_name_rule.dart
```diff
@@ -0,0 +1,49 @@
...
+
+void main() {
+  group('BanNameRule', () {
+    test('initialization', () async {
+      final unit = await RuleTestHelper.resolveFromFile(_examplePath);
+      final issues = BanNameRule().check(unit);
+
+      RuleTestHelper.verifyInitialization(
+        issues: issues,
+        ruleId: 'prefer-correct-type-name',
```
incendial, line 17: Interesting... and the test is not failing? Could you fix this, I'll take a look why it's not failing a bit later.

website/docs/rules/common/ban-name.md
```diff
@@ -0,0 +1,43 @@
+# Ban name
+
+## Rule id {#rule-id}
+
+ban-name
+
+## Severity {#severity}
+
+Warning
```
incendial, line 9: Probably should be `Style` instead.

website/docs/rules/common/ban-name.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Description {#description}
+
+Configure some names that you want to ban.
+
+For example, in my own app, I want to ban `showDialog` and force to use `myShowDialog` instead. This is because (let me explain the simplified case) I added logging in `myShowDialog`, and I do want every dialog to be logged.
```
incendial, line 15: Could you rephrase this description to a more abstract form? I totally get the idea and the example is legit, but it'd be better to have `you` as pointing to the user standpoint, not `I`.

lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/visitor.dart
```diff
@@ -0,0 +1,50 @@
...
+
+  @override
+  void visitLibraryIdentifier(LibraryIdentifier node) {
+    super.visitLibraryIdentifier(node);
+    for (final component in node.components) {
+      _visitIdent(node, component);
+    }
+  }
+
+  void _visitIdent(Expression node, SimpleIdentifier ident) {
```
incendial, line 34: Since it's an identifier, that means it will trigger for variable, method, function, class names, etc. (everything, that has an identifier). Have you thought about edge case when you can have a variable and a method called the same way (somehow) and triggering on one of them will be considered as a false-positive? Maybe there should be some config to limit the type of what contains that identifier?

incendial (commented):
Again, great result! Actually, I remember now that we had a request to add a rule to ban specific widget names (like, if you have a custom button and you want to ban the default one) and looks like your implementation covers this case too.

The main questions for me is about limiting the scoped of a configured name, like "trigger only on class names" or "trigger only on variables". What do you think about it, will this be useful?

fzyzcjy:
Thanks!

> The main questions for me is about limiting the scope of a configured name, like "trigger only on class names" or "trigger only on variables". What do you think about it, will this be useful?

I agree it may be quite useful. But I have not figured out an approach to do it easily. I have checked about visiting the function invocation node. But the "function name" is an Expression instead of a SimpleLiteral.

In addition, I guess few people will write down `var showDialog = ...` since that is like a *verb* not a *noun* so should be in a function. And they are always free to `// ignore: ban-name`.

Anyway, maybe firstly merge this PR, and later improve it in another PR when I have time?

dkrutskikh:
@fzyzcjy can you fix conflicts?

fzyzcjy:
> Totally agree, but for class properties and regular variables the amount or false positives might be higher. 

I may do that in a future PR :)

> Plus looks like this rule should not trigger on imported from external packages identifiers, since there will be no way to change them. Could you add this? I guess, that's the only blocker I see.

Could you please elaborate a bit more? For example, `showDialog` *is* from external packages, but I hope users never use it in *current* code.

> Also, could you please add more tests and examples to docs (like variables, class names, etc.), just to be sure, what works and what doesn't?

ok

fzyzcjy:
done

lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/visitor.dart
fzyzcjy: (discussed below)

fzyzcjy:
👀

fzyzcjy:
👀 👀

fzyzcjy:
A few days have passed, so I merge them into my forked repo's master branch. Further PR happens in https://github.com/dart-code-checker/dart-code-metrics/pull/722.

incendial:
Sorry for the long response, current political situation is a fcking disaster, will look at the changes ASAP

fzyzcjy:
Take your time. I have merged everything in my fork's `master` branch.

Issue #474, opened 2021-09-24
[New rule] prefer-immediate-return


roman-petrov:
**Please describe what the rule should do:**

The rule should disallow declaring variables just before returning them.

**If your rule is inspired by other please provide link to it:**

The rule is the same as [prefer-immediate-return](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/prefer-immediate-return.md) from `eslint-plugin-sonar-js`.

**What category of rule is this? (place an "X" next to just one item)**

[ ] Warns about a potential error (problem)
[x] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule will warn about (it will be better if you can provide both good and bad examples):**

See [SonarJS rule](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/prefer-immediate-return.md)

**Are you willing to submit a pull request to implement this rule?**
Maybe

incendial:
Available in 4.13 🚀

PR #736, opened 2022-03-11
feat: add prefer-immediate-return rule

roman-petrov:

[ ] Documentation update
[ ] Bug fix
[x] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

https://github.com/dart-code-checker/dart-code-metrics/issues/474

The fixer implemented is something like "semi-automatic" fixer, it does only half of the work required :) It replaces only expression in the problematic return statement and code contains unused variable after that.

For example,

```dart
int calculateSum(int a, int b) {
   final sum = a + b; 
   return sum;
}
```

will be fixed as 

```dart
int calculateSum(int a, int b) {
   final sum = a + b; 
   return a + b;
}
```

and `sum` variable should be manually removed. I think that this shoud be acceptable - it's not a huge amount of work to manually remove unused variable (linters should warn about it too).

Implementing fully automatic fixer seems to be a bit hard: we have to replace multiple nodes/statements, but currently (as I see) we have only `nodeLocation` utility method which can create location only from single node.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * **Breaking Change:** cli arguments `--fatal-unused` and `--fatal-warnings` activate by default.
 * chore: restrict `analyzer` version to `>=3.0.0 <3.4.0`.
 * chore: restrict `analyzer_plugin` version to `>=0.9.0 <0.10.0`.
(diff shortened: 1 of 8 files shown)


vlkonoshenko (approved):

incendial (approved):
🔥

dkrutskikh (approved):

PR #739, opened 2022-03-13
docs: Improve readability of no-equal-then-else description

rmortes:

[X] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

I changed "value1" and "value2" to "firstValue" and "secondValue" in the docs for no-equal-then-else.

This change may seen small, but I've been stuck reading the two same code blocks for around 5 minutes,
trying to figure out whether the docs were wrong or I was having an aneurysm. The numbers 1 and 2 are too similar
to be easily distinguisable, but "first" and "second" have different lengths, so it should be
easier for another devs to distiguish between the two variables

### DIFF
--- a/website/docs/rules/common/no-equal-then-else.md
+++ b/website/docs/rules/common/no-equal-then-else.md
@@ -17,38 +17,38 @@ Warns when if statement has equal then and else statements or conditional expres
 Bad:
 
 ```dart
-final value1 = 1;
-final value2 = 2;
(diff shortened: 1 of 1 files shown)


dkrutskikh:
@rmortes thank you for contribution!

Issue #754, opened 2022-03-22
[BUG] prefer-immediate-return does not work inside nested blocks


furaiev:
E.g. 

```
void calculateSum(int a, int b) {
    final sum = a + b;

    return sum; // LINT
}

void calculateSum(int a, int b) {
    try {
        final sum = a + b;

        return sum; // LINT DOES NOT WORK
    } ...
}
```

dkrutskikh:
@roman-petrov please take a look

roman-petrov:
@furaiev Thank you, good catch! This seems to be clear so I will make a PR in a few days.

incendial:
Available in 4.14.0 🚀

PR #760, opened 2022-03-25
fix: cover more cases in "prefer-immediate-return" rule

roman-petrov:
Fixes https://github.com/dart-code-checker/dart-code-metrics/issues/754

I just added more test cases and visitor now overrides `visitBlock` instead of overriding `visitBlockFunctionBody` to cover these cases in `prefer-immediate-return` rule.


roman-petrov:
"Auto Author Assign" check failed - seems this is not related to my PR so it should be ready for review.

dkrutskikh (approved):

incendial (approved):

PR #780, opened 2022-04-01
fix: correctly split exclude path

incendial:

[ ] Documentation update
[X] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/utils/exclude_utils.dart
+++ b/lib/src/utils/exclude_utils.dart
@@ -16,6 +16,5 @@ Iterable<Glob> prepareExcludes(
             Glob(p.normalize(p.join(root, exclude)).replaceAll(r'\', '/')))
         .toList();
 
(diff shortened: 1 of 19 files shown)


lib/src/utils/exclude_utils.dart
```diff
@@ -15,3 +15,7 @@ Iterable<Glob> prepareExcludes(
         .map((exclude) =>
             Glob(p.normalize(p.join(root, exclude)).replaceAll(r'\', '/')))
         .toList();
+
+Iterable<String> extractExcludes(String excludePath) => excludePath.isNotEmpty
```
dkrutskikh, line 19: @incendial can be simple

=> excludePath.split(',').map((path) => path.trim());

dkrutskikh (changes requested):

Issue #812, opened 2022-04-25
[BUG] `format-comment` is listing the macros from dart doc


ValentinVignal:
* **Dart code metrics version:**
4.14.0

* **Dart sdk version:**
```
Dart SDK version: 2.16.2 (stable) (Tue Mar 22 13:15:13 2022 +0100) on "macos_x64"
```

**Please show your full configuration:**

<details>
<summary>Configuration</summary>

```yaml
include: package:flutter_lints/flutter.yaml

analyzer:
  plugins:
    - dart_code_metrics
  strong-mode:
    implicit-casts: false
    implicit-dynamic: true
  errors:
    undefined_prefixed_name: ignore
linter:
  rules:
    always_put_required_named_parameters_first: true
    always_use_package_imports: true
    avoid_bool_literals_in_conditional_expressions: true
    avoid_classes_with_only_static_members: true
    avoid_double_and_int_checks: true
    avoid_dynamic_calls: true
    avoid_escaping_inner_quotes: true
    avoid_field_initializers_in_const_classes: true
    avoid_js_rounded_ints: true
    avoid_multiple_declarations_per_line: true
    avoid_null_checks_in_equality_operators: true
    avoid_private_typedef_functions: true
    avoid_redundant_argument_values: true
    avoid_returning_null_for_future: true
    avoid_setters_without_getters: true
    avoid_shadowing_type_parameters: true
    avoid_slow_async_io: true
    avoid_type_to_string: true
    avoid_types_on_closure_parameters: true
    avoid_unused_constructor_parameters: true
    avoid_void_async: true
    cancel_subscriptions: true
    cascade_invocations: true
    cast_nullable_to_non_nullable: true
    close_sinks: true
    comment_references: true
    conditional_uri_does_not_exist: true
    depend_on_referenced_packages: true
    directives_ordering: true
    empty_constructor_bodies: true
    eol_at_end_of_file: true
    flutter_style_todos: true
    invariant_booleans: true
    join_return_with_assignment: true
    leading_newlines_in_multiline_strings: true
    lines_longer_than_80_chars: false # We use 160: true
    literal_only_boolean_expressions: true
    missing_whitespace_between_adjacent_strings: true
    no_adjacent_strings_in_list: true
    no_logic_in_create_state: true
    no_runtimeType_toString: true
    noop_primitive_operations: true
    null_check_on_nullable_type_parameter: true
    omit_local_variable_types: true
[...]

PR #823, opened 2022-04-30
fix: `format-comment` is listing the macros from dart doc.

vlkonoshenko:

[ ] Documentation update
[X] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

https://github.com/dart-code-checker/dart-code-metrics/issues/812

### DIFF
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/format_comment/visitor.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/format_comment/visitor.dart
@@ -44,20 +44,31 @@ class _Visitor extends RecursiveAstVisitor<void> {
 
     var text = commentText.trim();
 
-    if (text.isEmpty ||
-        text.startsWith('ignore:') ||
-        text.startsWith('ignore_for_file:')) {
-      return;
-    } else {
-      text = text.trim();
-      final upperCase = text[0] == text[0].toUpperCase();
-      final lastSymbol = _punctuation.contains(text[text.length - 1]);
-      final hasEmptySpace = commentText[0] == ' ';
-      final incorrectFormat = !upperCase || !hasEmptySpace || !lastSymbol;
-      final single = commentToken.previous == null && commentToken.next == null;
+    final isIgnoreComment =
+        text.startsWith('ignore:') || text.startsWith('ignore_for_file:');
 
-      if (incorrectFormat && single) {
-        _comments.add(_CommentInfo(type, commentToken));
+    final regTemplateExp = RegExp(r"{@template [\w'-]+}");
+    final regMacroExp = RegExp(r"{@macro [\w'-]+}");
+
+    final isMacros = regTemplateExp.hasMatch(text) ||
+        regMacroExp.hasMatch(text) ||
+        text == '{@endtemplate}';
+
+    {
+      if (text.isEmpty || isIgnoreComment || isMacros) {
+        return;
+      } else {
+        text = text.trim();
+        final upperCase = text[0] == text[0].toUpperCase();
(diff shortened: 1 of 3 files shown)


incendial (approved):

lib/src/analyzers/lint_analyzer/rules/rules_list/format_comment/visitor.dart
```diff
@@ -44,20 +44,31 @@ class _Visitor extends RecursiveAstVisitor<void> {
...
-      final hasEmptySpace = commentText[0] == ' ';
-      final incorrectFormat = !upperCase || !hasEmptySpace || !lastSymbol;
-      final single = commentToken.previous == null && commentToken.next == null;
+    final isIgnoreComment =
+        text.startsWith('ignore:') || text.startsWith('ignore_for_file:');
 
-      if (incorrectFormat && single) {
-        _comments.add(_CommentInfo(type, commentToken));
+    final regTemplateExp = RegExp(r"{@template [\w'-]+}");
+    final regMacroExp = RegExp(r"{@macro [\w'-]+}");
```
ookami-kb, line 51: I would make them file-private final variables. It can look like a micro-optimization, but for analyzer running on thousands comments that could matter.

E.g. this benchmark:

```dart
import 'package:benchmark/benchmark.dart';

void main() {
  benchmark('Create RegExp', () {
    List.generate(100000, (i) {
      final x = RegExp(r"{@template [\w'-]+}");
      return x.hasMatch('abc');
    });
  });

  benchmark('Reise RegExp', () {
    List.generate(100000, (i) {
      return _regex.hasMatch('abc');
    });
  });
}

final _regex = RegExp(r"{@template [\w'-]+}");
```

gives such a result:

```
❯ dart run benchmark --dir=.
 DONE  ././lib/example_benchmark.dart (5 s)
 ✓ Create RegExp (103 ms)
 ✓ Reise RegExp (22 ms)

Benchmark suites: 1 passed, 1 total
Benchmarks:       2 passed, 2 total
Time:             5 s
Ran all benchmark suites.
```

incendial: Nice catch, actually!

lib/src/analyzers/lint_analyzer/rules/rules_list/format_comment/visitor.dart
```diff
@@ -5,6 +5,9 @@ const commentsOperator = {
   _CommentType.documentation: '///',
 };
 
+final _regTemplateExp = RegExp(r"{@template [\w'-]+}");
+final _regMacroExp = RegExp(r"{@macro [\w'-]+}");
+
```
ValentinVignal, line 10: Most of the example I've seen online also uses `.` in the template names. And this is supported by dartdoc.

Wouldn't this be better?

```suggestion
final _regTemplateExp = RegExp(r"{@template [\w'-\.]+}");
final _regMacroExp = RegExp(r"{@macro [\w'-\.]+}");

```

ex: 

```dart
/// {@template my_project.my_class.my_method}
```

ookami-kb: In general, it can even be combined and simplified to `r'{@(template|macro) .+}'`.

I don't see any actual profit in making it stricter.

vlkonoshenko: Thanks! Done))

dkrutskikh (approved):

PR #862, opened 2022-05-22
fix: improve context root included files calculation

incendial:

[ ] Documentation update
[X] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

### DIFF
--- a/lib/src/utils/analyzer_utils.dart
+++ b/lib/src/utils/analyzer_utils.dart
@@ -36,7 +36,7 @@ Set<String> getFilePaths(
     final newPath = normalize(join(rootFolder, path));
     final rootPath = context.contextRoot.root.path;
 
-    return newPath == rootPath || rootPath.startsWith('$newPath/');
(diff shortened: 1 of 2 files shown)


lib/src/utils/analyzer_utils.dart
```diff
@@ -36,7 +36,7 @@ Set<String> getFilePaths(
     final newPath = normalize(join(rootFolder, path));
     final rootPath = context.contextRoot.root.path;
 
-    return newPath == rootPath || rootPath.startsWith('$newPath/');
+    return newPath == rootPath || newPath.startsWith('$rootPath/');
```
dkrutskikh, line 39: @incendial please cover by unit test

dkrutskikh (changes requested):

lib/src/utils/analyzer_utils.dart
incendial: Done

Issue #799, opened 2022-04-14
[BUG] no errors reported when running on current directory (`.`)


DetachHead:
* **Dart code metrics version:** 4.13.0
* **Dart sdk version:** 2.16.2

**Please show your full configuration:**

<details>
<summary>Configuration</summary>

```yaml
analyzer:
  plugins:
    - dart_code_metrics
  language:
    strict-casts: true
    strict-inference: true
    strict-raw-types: true
  exclude:
    - lib/generated_plugin_registrant.dart
```

</details>

**What did you do? Please include the source code example causing the issue.**

```
>flutter pub run dart_code_metrics:metrics analyze .
✔ no issues found!

>flutter pub run dart_code_metrics:metrics analyze lib
lib\screens\report.dart:
STYLE   Block is empty. Empty blocks are often indicators of missing code.
        lib\screens\report.dart:85:29
        no-empty-block : https://dartcodemetrics.dev/docs/rules/common/no-empty-block

STYLE   Block is empty. Empty blocks are often indicators of missing code.
        lib\screens\report.dart:90:27
        no-empty-block : https://dartcodemetrics.dev/docs/rules/common/no-empty-block

STYLE   Block is empty. Empty blocks are often indicators of missing code.
        lib\screens\report.dart:104:29
        no-empty-block : https://dartcodemetrics.dev/docs/rules/common/no-empty-block
```

**What did you expect to happen?**
same errors when running in `.`

**What actually happened?**
no errors found

incendial:
@DetachHead have you tried calling with `./`? What's the output?

DetachHead:
```
>flutter pub run dart_code_metrics:metrics analyze ./
✔ no issues found!
```

incendial:
Available in 4.16.0! 🚀

DetachHead:
@incendial it doesn't seem to be fixed for me

```
>flutter pub run dart_code_metrics:metrics analyze .
✔ no issues found!

>flutter pub run dart_code_metrics:metrics analyze ./
✔ no issues found!

>flutter pub run dart_code_metrics:metrics analyze test
test\test.dart:
STYLE   Block is empty. Empty blocks are often indicators of missing code.
        test\test.dart:11:12
        no-empty-block : https://dartcodemetrics.dev/docs/rules/common/no-empty-block

>flutter pub run dart_code_metrics:metrics --version   
Dart Code Metrics version: 4.16.0
```

incendial:
[...]

PR #864, opened 2022-05-22
fix: normalize file paths after extraction from analyzed folder

incendial:

[ ] Documentation update
[X] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

#799

### DIFF
--- a/lib/src/utils/file_utils.dart
+++ b/lib/src/utils/file_utils.dart
@@ -23,5 +23,5 @@ Set<String> extractDartFilesFromFolders(
                   relative(entity.path, from: rootFolder),
                   globalExcludes,
                 ))
-            .map((entity) => entity.path))
+            .map((entity) => normalize(entity.path)))
         .toSet();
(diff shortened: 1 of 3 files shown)


lib/src/utils/file_utils.dart
```diff
@@ -23,5 +23,5 @@ Set<String> extractDartFilesFromFolders(
                   relative(entity.path, from: rootFolder),
                   globalExcludes,
                 ))
-            .map((entity) => entity.path))
+            .map((entity) => normalize(entity.path)))
```
dkrutskikh, line 26: @incendial please cover by tests

dkrutskikh (changes requested):

lib/src/utils/file_utils.dart
incendial: Done

PR #868, opened 2022-05-26
feat: add new rule `avoid-banned-imports` & modify `prefer-extracting-callbacks` allow short callbacks to exist

### LINKED ISSUE #820: [New rule] avoid-restricted-imports

I propose a rule similar to ESLint [no-restricted-imports](https://eslint.org/docs/rules/no-restricted-imports). There might be many use cases for it. In our project (a design system component library) we need to restrict import from Flutter `material.dart` because we do not use Material Design and must not depend on it anywhere.

Without configuration this rule should simply do nothing. I think we need to have a flexible configuration here, the most important for me is to configure rule per file glob pattern.

In nearest future I would like to discuss this rule and also I plan to provide a PR.

[no-restricted-imports](https://eslint.org/docs/rules/no-restricted-imports).

[x] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

Just possible rule configuration:

```yaml
- avoid_restricted_imports
   - paths: ["some/**/folder/*.dart", "another/folder/**.dart"]
     deny: ["package:flutter/material.dart"]
     message: "Do not import Flutter Material Design library, we should not depend on it!"
   - paths: ["core/*.dart"]
     deny: ["package:flutter_bloc/flutter_bloc.dart"]
     message: "State management should be not used inside "core" folder."
```

**BAD**

```dart
import "package:flutter/material.dart";
import "package:flutter_bloc/flutter_bloc.dart";
```

**GOOD**
```
No restricted imports in listed folders.
```

Yes, in a few weeks.

### LINKED ISSUE #866: [New rule] `forbid-import`, but allow `import ... show`

Hi thanks for the flutter-favorite package! In my app I have a layered structure, and some layer should not see some other layer. For example, a "domain store" layer should not know how UI looks like. 

Thus, I hope to forbid some kinds of imports in some files. For example, a config goes like:

```
dir: lib/domain_stores/**/*.dart
forbid_import:
- package:myapp/widgets/**
- package:myapp/pages/**
- package:flutter/** # because flutter package has UI things
```

Then, say `lib/domain_stores/store_one.dart` is:

```
import 'package:flutter/material.dart'; // LINT
import 'package:myapp/widgets/some_widget.dart'; // LINT
import 'package:myapp/other_things.dart'; // this is ok

... some code ...
```

[ ] Other (please specify:)

```dart

```

---

## EDIT: the `show` logic

It would be great to allow partial imports and forbid the other part. For example, in my domain store, I do not want it to know flutter `Widget`s, surely, as that is a part of UI. However, I do accept that it uses `MaterialColor`, since "colors" is less related to UI indeed.

Thus, I hope the following could be created:

```
dir: lib/domain_stores/**/*.dart
forbid_import:
- path: package:flutter/material.dart
  whitelistShow: [MaterialColor, Color, ...]
- path: package:flutter/**
- path: something_else
...
```

Then,

```
import 'package:flutter/material.dart'; // LINT
import 'package:myapp/widgets/some_widget.dart'; // LINT
import 'package:myapp/other_things.dart'; // this is ok

import 'package:flutter/material.dart' show MaterialColor; // THIS IS OK AS WELL!
```

fzyzcjy:
#820
#866
#873

[ ] Documentation update
[ ] Bug fix
[x] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:


fzyzcjy:
p.s. ignore the long list of commits above. they are from old merge history, and is unrelated to this pr.

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_banned_imports/utils/config_parser.dart
```diff
@@ -0,0 +1,40 @@
...
+
+        return _AvoidBannedImportsConfigEntry(
+          paths: _parseListRegExp(entryMap[_pathsLabel]),
+          deny: _parseListRegExp(entryMap[_denyLabel]),
+          message: entryMap[_messageLabel] as String,
+        );
+      }).toList();
+
+  static List<RegExp> _parseListRegExp(Object? object) =>
+      (object! as List<Object?>)
```
incendial, line 24: Those casts and null assertions look fragile to me. Is it possible to rewrite it with `as` and `whereType`?

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_banned_imports/visitor.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  final _nodes = <_NodeWithMessage>[];
+
+  Iterable<_NodeWithMessage> get nodes => _nodes;
+
+  _Visitor(this._activeEntries);
+
+  @override
+  void visitImportDirective(ImportDirective node) {
+    // print('hi node=$node uri=${node.uri} uri.stringValue=${node.uri.stringValue}');
```
incendial, line 14: Please remove this commented code

incendial (commented):
Left few small comments, everything else looks great

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_banned_imports/utils/config_parser.dart
fzyzcjy: IMHO fail fast is important. If the user provides wrong types it should immediately fail.

I just mimic `json_serializable`'s auto gen code.

incendial: Have you tested it with IDE highlights? I'm not sure, that plugin will be recreated on the config change. Could you check that?
For example, enable the rule in other repository referencing local copy, set the wrong config and then set the right one. If the plugin starts showing the errors for this rule - everything is fine

incendial: But in general I agree with fail fast approach, but I not sure that plugins are 😄

fzyzcjy: I guess we can see it at least in the analysis server's webpage. No bandwidth to try it though.

incendial: Hm, I actually noticed that saving changes to analysis_options leads to the analyzer restart. I think it might work fine.

incendial:
@roman-petrov do you want to participate in the review?

roman-petrov:
> @roman-petrov do you want to participate in the review?

I think that this is not necessary: I already reviewed the implementation and I think it covers majority of cases we need to start with.

fzyzcjy:
I am going to implement other features... So if this is not merged, two features will be in one single Pr.

fzyzcjy:
ready for review

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * test: added test case in [`prefer-const-border-radius`](https://dartcodemetrics.dev/docs/rules/flutter/prefer-const-border-radius) rule.
 * chore: restrict `analyzer` version to `>=2.4.0 <4.2.0`.
 * fix: improve context root included files calculation.
+* feat: add [`avoid-banned-imports`](https://dartcodemetrics.dev/docs/rules/common/avoid-banned-imports) rule
```
incendial, line 8: Please update the changelog with prefer-extracting-callbacks changes

lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_extracting_callbacks/prefer_extracting_callbacks_rule.dart
```diff
@@ -6,6 +6,7 @@ import 'package:analyzer/dart/ast/visitor.dart';
 import '../../../../../utils/flutter_types_utils.dart';
 import '../../../../../utils/node_utils.dart';
 import '../../../lint_utils.dart';
+import '../../../metrics/metrics_list/source_lines_of_code/source_code_visitor.dart';
```
incendial, line 9: Although I understand the idea of using the existing visitor, the general approach is to not import parts of modules into one another (like linter rules and metrics are separated), I'm not really sure how it should be change the best way. Maybe by moving the visitor to the common root folder?

lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_extracting_callbacks/visitor.dart
```diff
@@ -74,4 +81,16 @@ class _InstanceCreationVisitor extends RecursiveAstVisitor<void> {
...
       argument is! NamedExpression ||
       !_ignoredArguments.contains(argument.name.label.name);
+
+  bool _isLongEnough(Expression expression) {
+    final allowedLineCount = _allowedLineCount;
+    if (allowedLineCount == null) {
+      return true;
+    }
+
+    final visitor = SourceCodeVisitor(_source.lineInfo);
```
incendial, line 91: Since you need only the lineInfo, could you pass just it?

incendial (commented):
Sorry for the long delay, left a few small comments. Also, could you add ignores for the dependency_validator as listed here https://pub.dev/packages/dependency_validator#configuration ?

incendial:
@fzyzcjy will you be able to take a look at the comments above? if not, it's totally okay, I can address them myself (you already did a great job!)

fzyzcjy:
Oh I forgot it... I have been busy recently so maybe have no time to fix the minor issues above.

incendial:
> Oh I forgot it... I have been busy recently so maybe have no time to fix the minor issues above.

Sure, will fix it myself then

PR #874, opened 2022-05-28
feat: introduce new command check-unnecessary-nullable

incendial:

[ ] Documentation update
[ ] Bug fix
[ ] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[X] Other, please explain: new command

New command to check if a parameter of a function, a method or a constructor is marked nullable, but used everywhere as non-nullable. Removing nullability from such parameters can remove additional nullable checks and reduce the complexity of  code.


lib/src/analyzers/unnecessary_nullable_analyzer/invocations_visitor.dart
```diff
@@ -0,0 +1,56 @@
...
+
+    _recordUsedElement(node.constructorName.staticElement, node.argumentList);
+  }
+
+  /// Records use of a not prefixed [element].
+  void _recordUsedElement(Element? element, ArgumentList arguments) {
+    if (element == null) {
+      return;
+    }
+    // // Ignore if an unknown library.
```
dkrutskikh, line 48: @incendial strange comment

dkrutskikh (approved):

PR #918, opened 2022-07-12
feat: add avoid-shrink-wrap-in-lists

incendial:

[ ] Documentation update
[ ] Bug fix
[X] New rule
[ ] Changes an existing rule
[ ] Add autofixing to a rule
[ ] Add a CLI option
[ ] Add something to the core
[ ] Other, please explain:

Implementing https://github.com/dart-lang/linter/issues/3496

### DIFF
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_shrink_wrap_in_lists/visitor.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_shrink_wrap_in_lists/visitor.dart
@@ -1,0 +1,33 @@
+part of 'avoid_shrink_wrap_in_lists_rule.dart';
+
+class _Visitor extends RecursiveAstVisitor<void> {
+  final _expressions = <Expression>[];
+
+  Iterable<Expression> get expressions => _expressions;
+
+  @override
+  void visitInstanceCreationExpression(InstanceCreationExpression node) {
+    super.visitInstanceCreationExpression(node);
+
+    if (isListViewWidget(node.staticType) &&
+        _hasShrinkWrap(node) &&
+        _hasParentList(node)) {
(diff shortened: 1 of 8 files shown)


lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_shrink_wrap_in_lists/visitor.dart
```diff
@@ -0,0 +1,33 @@
...
+    }
+  }
+
+  bool _hasShrinkWrap(InstanceCreationExpression node) =>
+      node.argumentList.arguments.firstWhereOrNull(
+        (arg) => arg is NamedExpression && arg.name.label.name == 'shrinkWrap',
+      ) !=
+      null;
+
+  bool _hasParentList(InstanceCreationExpression node) =>
```
gabrielaraujoz, line 25: hey @incendial , awesome work! I took the liberty of reviewing your PR and I think there's one way to make it better: when we have a list inside another like we're testing, we usually have to wrap them in an Expanded Widget (check the bad case scenario). Because of that, I believe it would be better if we test for that case as well in this rule.

WDYT? Does that make sense to you? Let me know if you need me to make the adjustments for you, I'd be glad to help!

dkrutskikh (approved):

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_shrink_wrap_in_lists/visitor.dart
incendial: @gabrielaraujoz but it looks like wrapping in Expanded is a subset of all cases when a ListView is in the Column/Row, isn't it? Meaning, that the current check already covers it

Issue #944, opened 2022-07-26
[BUG] avoid-top-level-members-in-tests in lib


lsaudon:
* **Dart code metrics version:** 4.17.0
* **Dart sdk version:** 2.17.6

**Please show your full configuration:**

<details>
<summary>Configuration</summary>

```yaml
include: package:very_good_analysis/analysis_options.3.0.1.yaml

analyzer:
  plugins:
    - dart_code_metrics

linter:
  rules:
    public_member_api_docs: false
    sort_constructors_first: false

dart_code_metrics:
  rules-exclude:
    - .dart_tool
  rules:
    - avoid-banned-imports
    - avoid-collection-methods-with-unrelated-types
    - avoid-duplicate-exports
    - avoid-dynamic:
      exclude:
        - app_bloc_observer.dart
    - avoid-global-state
    - avoid-ignoring-return-values
    - avoid-late-keyword:
      exclude:
        - test/**
    - avoid-missing-enum-constant-in-map
    - avoid-nested-conditional-expressions
    - avoid-non-ascii-symbols
    - avoid-non-null-assertion
    - avoid-throw-in-catch-block
    - avoid-top-level-members-in-tests
    - avoid-unnecessary-type-assertions
    - avoid-unnecessary-type-casts
    - avoid-unrelated-type-assertions
    - avoid-unused-parameters
    - ban-name
    - binary-expression-operand-order
    - double-literal-format
    - format-comment
    # - member-ordering
    - member-ordering-extended
    - newline-before-return
    - no-boolean-literal-compare
    - no-empty-block
    # - no-equal-arguments
    - no-equal-then-else
    - no-magic-number
    - no-object-declaration
    - prefer-async-await
    - prefer-commenting-analyzer-ignores
    - prefer-conditional-expressions
    - prefer-correct-identifier-length
    - prefer-correct-type-name
    - prefer-enums-by-name
    - prefer-first
    - prefer-immediate-return
    - prefer-last
    - prefer-match-file-name:
      exclude:
        - l10n/**
    - prefer-trailing-comma
    - tag-name
    - always-remove-listener
    - avoid-border-all
    - avoid-returning-widgets
    - avoid-unnecessary-setstate
    - avoid-wrapping-in-padding
    - avoid-shrink-wrap-in-lists
    - avoid-use-expanded-as-spacer
    - prefer-const-border-radius
    - prefer-correct-edge-insets-constructor
    - prefer-extracting-callbacks
[...]

PR #945, opened 2022-07-26
fix: avoid-top-level-members-in-tests

lsaudon:

- [ ] Documentation update
- [x] Bug fix
- [ ] New rule
- [ ] Add a CLI option
- [ ] Other, please explain:

Issues: https://github.com/dart-code-checker/dart-code-metrics/issues/944

I added all the folders (`'*'`) so that only the test folder is taken into account.

I tried to do a test to see if the file was being ignored, but I have the impression that the exclusion is done in another part of the code.

As it stands, this is the only rule that doesn't take excludes in parameters.
Is this normal?
We could add the excludes of the config or replace the existing excludes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * chore: restrict `analyzer` version to `>=4.1.0 <4.4.0`.
 * chore: restrict `analyzer_plugin` version to `>=0.11.0 <0.12.0`.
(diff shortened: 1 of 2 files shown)


lsaudon:
I don't know why it was `Publish test results` that failed.

incendial (approved):

PR #987, opened 2022-09-04
feat: avoid passing async when sync expected

### LINKED ISSUE #886: [New rule] Warn when returning Future<void> instead of Void
Should give me an error when passing an async void function in place of a sync void function.

- https://stackoverflow.com/questions/72529983/flutter-how-to-detect-async-code-in-a-ui-callback/

[x ] Warns about a potential error (problem)
[ ] Other (please specify:)

```dart
typedef VoidCallback = void Function();

void hello(VoidCallback foo) {
  foo();
  print("hello dodo");
}

Future<void> cb() async {
  await Future.delayed(Duration(seconds: 1));
  print('Callback');
}

void main() {
  // Lets get an error when passing the return type 
  // doesn't match exactly
  hello(cb);
}
```

```dart
RaisedButton(
  child: Text("Next Page"),
  onPressed: () async {
    // This could be any long running task
    // The user can now press this button multiple times
    // and each subsequent press will result in an error
    // as the context won't be valid any more.
    await Future.delayed(Duration(seconds: 1));
    Navigator.pop(context);
  }),
),
```

**Are you willing to submit a pull request to implement this rule?** Maybe. But I don't want to commit to anything.

Ascenio:

- [ ] Documentation update
- [ ] Bug fix
- [x] New rule
- [ ] Add a CLI option
- [ ] Other, please explain:

I've added a new rule, `avoid-passing-async-when-sync-expected`, which warns the user about async functions being passed as argument where a sync function was expected. This is helpful because the function won't be awaited as the user may expect.

#886

Please, feel free to give advice on any other way to write the same rule in a more elegant way. I'm not familiar with the analyzer API xD.

### DIFF
--- a/lib/src/analyzers/lint_analyzer/rules/rules_factory.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_factory.dart
@@ -13,6 +13,7 @@ import 'rules_list/avoid_missing_enum_constant_in_map/avoid_missing_enum_constan
 import 'rules_list/avoid_nested_conditional_expressions/avoid_nested_conditional_expressions_rule.dart';
 import 'rules_list/avoid_non_ascii_symbols/avoid_non_ascii_symbols_rule.dart';
 import 'rules_list/avoid_non_null_assertion/avoid_non_null_assertion_rule.dart';
+import 'rules_list/avoid_passing_async_when_sync_expected/avoid_passing_async_when_sync_expected.dart';
 import 'rules_list/avoid_preserve_whitespace_false/avoid_preserve_whitespace_false_rule.dart';
 import 'rules_list/avoid_returning_widgets/avoid_returning_widgets_rule.dart';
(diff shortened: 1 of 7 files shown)


incendial:
@Ascenio this looks amazing, you nailed it! 🚀 

Please update the changelog and I'm ready to merge.

incendial:
Looks like the test doesn't pass:

Screenshot 2022-09-11 at 18 05 27

NANASHI0X74:
This seems to throw warnings when it shouldn't, particularly for Future's then<R>(FutureOr<R> onValue(T value), {Function? onError}) and similarly for whenComplete and flutter's Future.onError methods- example:
```dart
  Future<String> doSomeGetRequest() => Future.value('');

  Future<String> doAnotherGetRequest(String input) => Future.value('');

  Future<String> main() => doSomeGetRequest().then(doAnotherGetRequest); // lint
```
throws a lint, but then() accepts FutureOr, so asynchronous code is expected there

incendial:
@NANASHI0X74 could you please file a separate issue with the problem?

PR #988, opened 2022-09-04
fix: make avoid-border-all not report errors on final variables

### LINKED ISSUE #973: [BUG] avoid-border-all is triggered even when it is not a const
There is a regression which is exactly https://github.com/dart-code-checker/dart-code-metrics/issues/837. The rule `avoid-border-all` is triggered even when the parameters are not constants.

* **Dart code metrics version:**
`4.18.0-dev.2`
* **Dart sdk version:**
`2.17.6`

**Please show your full configuration:**

<details>
<summary>Configuration</summary>

```yaml
include: package:flutter_lints/flutter.yaml

analyzer:
  plugins:
    - dart_code_metrics

dart_code_metrics:
  rules:
    - avoid-border-all
```

</details>

**What did you do? Please include the source code example causing the issue.**

You can also checkout https://github.com/ValentinVignal/flutter_app_stable/tree/dart-code-metrics/border-and-radius

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    final boolean = MediaQuery.of(context).size.width > 300;
    final color = boolean ? Colors.red : Colors.black;
    return MaterialApp(
      home: Container(
        decoration: BoxDecoration(
          border: Border.all(color: color), // <- Lint
        ),
      ),
    );
  }
}

```

**What did you expect to happen?**

I expect to see no lint when the variables inside the border are not constant

**What actually happened?**

I see a lint

```
Prefer using const constructor Border.fromBorderSide.
const Border.fromBorderSide(BorderSide(color: Theme.of(context).colorScheme.primary)) dart[avoid-border-all](https://dartcodemetrics.dev/docs/rules/flutter/avoid-border-all).
```

**Are you willing to submit a pull request to fix this bug?**

With help, I could

Ascenio:

- [ ] Documentation update
- [x] Bug fix
- [ ] New rule
- [ ] Add a CLI option
- [ ] Other, please explain:

I've made the following code not trigger warnings:
```dart
final aLocalColor = Colors.red;
Border.all(color: aLocalColor);
```

#973.

I would like to know If there's any way that we could refactor this rule such that we didn't have to cover each and every case separately.

### DIFF
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_border_all/visitor.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_border_all/visitor.dart
@@ -29,6 +29,8 @@ class _Visitor extends RecursiveAstVisitor<void> {
           final element = arg.staticElement;
           if (element is PropertyAccessorElement && !element.variable.isConst) {
             isAllConst = false;
+          } else if (element is VariableElement && !element.isConst) {
+            isAllConst = false;
           }
         }
       }
--- a/test/src/analyzers/lint_analyzer/rules/rules_list/avoid_border_all/avoid_border_all_rule_test.dart
+++ b/test/src/analyzers/lint_analyzer/rules/rules_list/avoid_border_all/avoid_border_all_rule_test.dart
@@ -5,6 +5,8 @@ import 'package:test/test.dart';
 import '../../../../../helpers/rule_test_helper.dart';
 
 const _examplePath = 'avoid_border_all/examples/example.dart';
+const _exampleWithVariablesPath =
+    'avoid_border_all/examples/example_with_variables..dart';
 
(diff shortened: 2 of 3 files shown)


incendial:
@Ascenio thank you so much for the fix and sorry for the delay. Could you please also update the changelog?

> I would like to know If there's any way that we could refactor this rule such that we didn't have to cover each and every case separately.

That's a very good question, @vlkonoshenko any thoughs?

vlkonoshenko:
@incendial I don't know another solution now. Because in first case we should take variable from node and in this case we take immediately a variable. I not found same interfaces in them

Issue #997, opened 2022-09-14
[New rule] check-for-equals-in-render-object-setters when implementing a RenderObject; maybe even apply this linter to check potential bugs in Flutter source code


fzyzcjy:
**Please describe what the rule should do:**

As long as we write down a few `RenderObject`s, or read through some in flutter source code, we will realize a pattern:

```
  set a(A a) {
    if (_a == a) return; // <---- NOTE
    _a = a;
    markNeedsPaint(); // or something else etc
  }
```

That early return is very helpful, because `set a` is always called within updateRenderObject. If it does not early return, then the markNeedsPaint (or for other fields, maybe markNeedsLayout, some heavy work, etc), will be *unconditionally* executed every time the render object is updated, even if that is totally unnecessary. 

Thus, this rule should find out such problem and lint it.

Btw, maybe we can even apply this linter to check potential bugs in Flutter source code.

**If your rule is inspired by other please provide link to it:**

**What category of rule is this? (place an "X" next to just one item)**

- [x] Warns about a potential error (problem)
- [ ] Suggests an alternate way of doing something (suggestion)
- [ ] Other (please specify:)

**Provide 2-3 code examples that this rule will warn about (it will be better if you can provide both good and bad examples):**

```dart
class A extends RenderBox {
  set a(A a) {
    if (_a == a) return; // <---- if no such line, should LINT
    _a = a;
    markNeedsPaint(); // or something else etc
  }
}
```

**Are you willing to submit a pull request to implement this rule?**

incendial:
@fzyzcjy could you share an example from Flutter codebase?

incendial:
I look at this one https://github.com/flutter/flutter/blob/5bc4a7612f7fd430933b046aeb9e36a5c7f3916e/packages/flutter/lib/src/cupertino/text_selection_toolbar.dart#L667 and it looks too hard to extract a formal rule coz of `_backButton = _updateChild` calls.

incendial:
The info I'm looking for: which classes should be checked and which setters should have the check.

incendial:
Right now it looks like it's `a class, that extends RenderObject` and `all setters`, but the example above breaks this rule

fzyzcjy:
@incendial 

> could you share an example from Flutter codebase?

Again, any RenderBox's subclass. For example, RenderPadding, RenderOpacity, RenderAnythingYouLike.

incendial:
[...]

PR #1003, opened 2022-09-16
feat: add static code diagnostic check-for-equals-in-render-object-setters

incendial:

- [ ] Documentation update
- [ ] Bug fix
- [X] New rule
- [ ] Add a CLI option
- [ ] Other, please explain:

#997

### DIFF
--- a/test/src/analyzers/lint_analyzer/rules/rules_list/check_for_equals_in_render_object_setters/examples/example.dart
+++ b/test/src/analyzers/lint_analyzer/rules/rules_list/check_for_equals_in_render_object_setters/examples/example.dart
@@ -1,0 +1,77 @@
+class RenderObject {}
+
+class RenderBox extends RenderObject {}
+
+class SomeRenderBox extends RenderBox {
+  int _page;
+  int get page => _page;
+  set page(int value) {
+    if (value == _page) {
+      return;
+    }
+    _page = value;
+    markNeedsLayout();
+  }
+
+  double get overflowSpacing => _overflowSpacing;
+  double _overflowSpacing;
+  set overflowSpacing(double value) {
+    assert(value != null);
+    if (_overflowSpacing == value) return;
+
+    _overflowSpacing = value;
+    markNeedsLayout();
+  }
+
+  double _dividerWidth;
+  double get dividerWidth => _dividerWidth;
+  // LINT
+  set dividerWidth(double value) {
+    _dividerWidth = value;
+    markNeedsLayout();
+  }
+
+  double _dividerHeight;
+  double get dividerHeight => _dividerHeight;
+  // LINT
+  set dividerHeight(double value) {
+    if (_dividerHeight == _dividerWidth) {
+      return;
+    }
+
+    _dividerHeight = value;
+    markNeedsLayout();
+  }
+
+  Clip get clipBehavior => _clipBehavior;
+  Clip _clipBehavior = Clip.none;
+  set clipBehavior(Clip value) {
+    assert(value != null);
+    if (value == _clipBehavior) {
+      return;
+    }
+    _clipBehavior = value;
+    markNeedsPaint();
+    markNeedsSemanticsUpdate();
+  }
+
+  double get spacing => _spacing;
+  double _spacing;
+  // LINT
+  set spacing(double value) {
+    _spacing = value;
+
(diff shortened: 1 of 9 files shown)


incendial:
@fzyzcjy if you have some time to take a look and test the rule - that'd very much appreciated!

fzyzcjy:
Sure, let me have a look :)

As for testing, may not have time now, but you can run it on the `flutter/flutter` repo, and even possibly fire a bug :)

test/src/analyzers/lint_analyzer/rules/rules_list/check_for_equals_in_render_object_setters/examples/example.dart
```diff
@@ -0,0 +1,77 @@
...
+    }
+    _page = value;
+    markNeedsLayout();
+  }
+
+  double get overflowSpacing => _overflowSpacing;
+  double _overflowSpacing;
+  set overflowSpacing(double value) {
+    assert(value != null);
+    if (_overflowSpacing == value) return;
```
fzyzcjy, line 20: IIRC there is also a case:

```dart
if (_overflowSpacing != value) {
  _overflowSpacing = value;
  markNeedsLayout()
}
```

But forget whether this is rare or not. Maybe we can run it on flutter framework repo first and see

fzyzcjy:
LGTM especially on the demo in the tests. But not sure how many false positives will happen in the real codebase - may need to run it actually :)

test/src/analyzers/lint_analyzer/rules/rules_list/check_for_equals_in_render_object_setters/examples/example.dart
incendial: Hm, thanks, looks valid

incendial:
Mostly false-positives with the case you pointed out before, but those:
* https://github.com/flutter/flutter/blob/7fba37848c9f3007795be2d67a3207067b7a0894/packages/flutter/lib/src/rendering/platform_view.dart#L323
* https://github.com/flutter/flutter/blob/8b3aabd61dfd7b7b525ae4be04093ba507d1455a/packages/flutter/lib/src/rendering/editable.dart#L596
* https://github.com/flutter/flutter/blob/a7fe235c8cd1ff64018443b0cca9a19a751642a6/packages/flutter/lib/src/rendering/paragraph.dart#L132
* https://github.com/flutter/flutter/blob/d3bc2bbcdd814beb343719d9b7fed58b3807675f/packages/flutter/lib/src/material/range_slider.dart#L891

look interesting. WDYT?

fzyzcjy:
> https://github.com/flutter/flutter/blob/a7fe235c8cd1ff64018443b0cca9a19a751642a6/packages/flutter/lib/src/rendering/paragraph.dart#L132

No problem since it "return" when "idential"

Checking the rest 3

fzyzcjy:
> https://github.com/flutter/flutter/blob/7fba37848c9f3007795be2d67a3207067b7a0894/packages/flutter/lib/src/rendering/platform_view.dart#L323

I strongly suspect this is a bug! That is what I have said in the issue title - "maybe even apply this linter to check potential bugs in Flutter source code"!

incendial:
> What about giving a PR to flutter repo fixing this 🎉

Do you want to do it? It's mostly your effort finding this 🙂

fzyzcjy:
> Do you want to do it?

Sure 🙂

incendial:
@fzyzcjy found another place https://github.com/flutter/flutter/blob/2aa348b9407e96ffe4eca8e8f213c7984afad3f7/packages/flutter/lib/src/widgets/sliver_prototype_extent_list.dart#L135, WDYT?

fzyzcjy:
@incendial Not very sure. The only interesting call is:

image

However it is named "insertRenderObjectChild", where by saying "insert" I suspect it will not happen unconditionally on every update. So I guess it is normal.

test/src/analyzers/lint_analyzer/rules/rules_list/check_for_equals_in_render_object_setters/examples/example.dart
incendial: Fixed, ~10 false-positives left, but I think they're mostly for manual ignore

Issue #927, opened 2022-07-19
[Rule change] Let `ban-name` ban things like `DateTime.now`


fzyzcjy:
To allow testability, it would be great to ban all `DateTime.now` and use `clock.now` instead. Thus, we can extend `ban-name` to do so.

**What rule do you want to change?**

**How will the change be implemented? (New option, new default behavior, etc.)?**

**Please provide some example code that this change will affect:**

```dart
```

**What does the rule currently do for this code?**

**What will the rule do after it's changed?**

**Are you willing to submit a pull request to implement this change?**

incendial:
Available in 5.1.0 🚀

fzyzcjy:
🎉

PR #1034, opened 2022-10-13
feat: add method call chains support on ban-name ident

Hideart:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New rule
- [x] Changes an existing rule
- [ ] Add a CLI option
- [ ] Other, please explain:

This PR extends the ban-name rule abilities to recursive search in method call chains. (ex: `DateTime.now` or `StrangeClass.someField.subField` etc.)
Ban-name rule's visitor will have a breadcrumb now and will compare breadcrumb node names with the rule ident.

Resolves the #927 

> **IMPORTANT:**
> Tests passes, CLI works as expected. But the IDE does not highlight method call chains. I'm trying to figure out what's wrong
>
> **UPD:**
> **Progress is [here](https://github.com/dart-lang/sdk/issues/50209)**

### DIFF
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/ban_name_rule.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/ban_name_rule.dart
@@ -33,7 +33,24 @@ class BanNameRule extends CommonRule {
 
     source.unit.visitChildren(visitor);
 
-    return visitor.nodes
+    final filteredNodeList = visitor.nodes.where((node) {
+      if (node.endNode != null) {
+        final inlineNode = source.content
+            .substring(
+              node.node.offset,
+              node.endNode!.end,
+            )
+            .replaceAll('\n', '');
+
+        if (!inlineNode.contains(node.fullName)) {
+          return false;
+        }
+      }
+
+      return true;
+    });
+
+    return filteredNodeList
         .map(
           (node) => createIssue(
             rule: this,
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/ban_name_rule.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/ban_name_rule.dart
@@ -37,7 +37,11 @@ class BanNameRule extends CommonRule {
         .map(
           (node) => createIssue(
             rule: this,
-            location: nodeLocation(node: node.node, source: source),
+            location: nodeLocation(
+              node: node.node,
+              source: source,
+              endNode: node.endNode,
+            ),
             message: node.message,
           ),
         )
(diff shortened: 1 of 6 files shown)


lib/src/analyzers/lint_analyzer/rules/rules_list/ban_name/ban_name_rule.dart
```diff
@@ -33,11 +33,32 @@ class BanNameRule extends CommonRule {
 
     source.unit.visitChildren(visitor);
 
-    return visitor.nodes
+    final filteredNodeList = visitor.nodes.where((node) {
+      if (node.endNode != null) {
```
incendial, line 37: Could you elaborate, what you're actually trying to check here?

Hideart: The visitor finds 2 adjacent nodes separated by a single character. This character seems to be only checkable inside a rule based on source.content
Without a filter, the last line in the test will be lint when it shouldn't be.

This filter trying to find an ident in range between `node.offset` and `endNode.end` inside a `source.content`
For ex.:
```
// analysis_options.yaml
... 
- ban-name:
     entries:
       - ident: DateTime.now
...

// code
DateTime now = DateTime(2022);
```
1. Without this filter, the visitor finds a `DateTime now` because it doesn't know anything about the separator
2. Filter checks if "node" `DateTime now = DateTime(2022);` contains an `DateTime.now` ident

incendial: Okay, I've cloned the implementation and tested it locally. 

I think the approach should be changed.
Instead of trying to build a string from identifier names, let's try other thing - let's check for node parent and if there is a method invocation, variable access or property access, we will extract the "path" from parent and check if it exists in the config.

So, if I add to the test `DateTime.now().day.isFinite;` and visit `isFinite`, the parent property will be the following:
Screenshot 2022-10-16 at 13 34 56

Changing the approach will allow to remove this check and check everything inside the visitor, plus will not rely on the string representation, but rather on parsed AST.

test/src/analyzers/lint_analyzer/rules/rules_list/ban_name/examples/example.dart
```diff
@@ -15,3 +16,13 @@ void strangeName() {} // LINT
...
+
+StrangeClass.someMethod(); // LINT
+NonStrangeClass.someMethod();
+
+DateTime.now(); // LINT
+DateTime.now().day; // LINT
+class DateTimeTest {
+  final currentTimestamp = DateTime.now(); // LINT
+}
+DateTime now = DateTime(2022);
```
incendial, line 28: Please wrap all the statements inside a main. I'm not sure that this code is parsed properly.

incendial (changes requested):

incendial:
@Hideart I'll merge this as is and fix myself. Thank you for the contribution!

Issue #904, opened 2022-06-29
[Feature request] cli command to fix all


LogicNg:
Hi all,

First off, I would like to say that this package is awesome. Thanks for all the effort you guys put in. It would be nice if we can further automate the process. For example, when I enable dart-code-checker in a project, I have to go through each warning and click the auto-fix button.

I'm aware that fix all on save is [currently impossible](https://github.com/dart-lang/sdk/issues/46760). As a workaround, can we make an CLI command to fix all warnings?

incendial:
@LogicNg hi, thank you! As for auto-fix, it is already in progress, hope it'll be ready in the next major release, but no exact date yet.

LogicNg:
@incendial Glad to hear that!

BrutalCoding:
Waiting eagerly for this solution! I have to go through a 1000 prefer-comma-trailing fixes manually now so hopefully there's some news about this CLI feature before I get it down to zero 😆

PR #1047, opened 2022-10-21
feat: add static code diagnostic arguments-ordering

roman-petrov:
## `arguments-ordering` rule

Enforces named argument order in function and constructor invocations to be the same as corresponding named parameter declaration order.

The rule can be used together with [sort_child_properties_last](https://dart-lang.github.io/linter/lints/sort_child_properties_last.html) rule by setting `child-last: true` in rule configuration.

### ⚙️ Config example

```yaml
dart_code_metrics:
  ...
  rules:
    ...
    - arguments-ordering:
        child-last: true
```

### Example

**❌ Bad:**

```dart
Person buildPerson({ String name, String surname, String age });
// LINT
final person = buildPerson(age: 42, surname: 'The Pooh', name: 'Winnie');
```

**✅ Good:**

```dart
Person buildPerson({ String name, String surname, String age });
// LINT
final person = buildPerson(name: 'Winnie', surname: 'The Pooh', age: 42);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* feat: add static code diagnostic [`arguments-ordering`](https://dartcodemetrics.dev/docs/rules/common/arguments-ordering).
+
 ## 5.0.1
 
 * fix: correctly available check rule names.
--- a/lib/src/analyzers/lint_analyzer/rules/rules_factory.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_factory.dart
@@ -1,5 +1,6 @@
 import 'models/rule.dart';
 import 'rules_list/always_remove_listener/always_remove_listener_rule.dart';
+import 'rules_list/arguments_ordering/arguments_ordering_rule.dart';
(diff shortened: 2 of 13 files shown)


roman-petrov:
I finally finished this PR :)

This rule might be a bit hard to adopt for large code bases without [CLI "auto-fix"](https://github.com/dart-code-checker/dart-code-metrics/issues/904). However, I managed to test this rule on our relatively large code base and spent just about half a hour to fix all ~250 issues in VSCode. Just monkey clicking "Fix" light bulb in problems panel:) 

I didn't found any issues with this rule on our code base, auto fix also worked pretty well. So I assume this implementation should be quite ready to ship.

 @dkrutskikh , @incendial  - could you please take a look? 

I've got a strange issue with automated checks for this PR and don't have any idea how it can be fixed:
image

incendial (approved):
Looks great 🔥 

But I bet someone might want to have a specific order 😁

roman-petrov:
> But I bet someone might want to have a specific order 😁
Sure 😁 But it will be quite easy to implement new sorting options having "core" rule part completed.

PR #1120, opened 2022-12-24
feat: add use-setstate-synchronously rule

### LINKED ISSUE #1117: [New rule] use-setstate-synchronously

(Flutter-specific) Attempts to call `setState` after the state has been disposed usually represents a potential bug. To remedy this, the user needs to either:
1. Insert a check for `mounted` and call `setState` in it
2. Insert a guard for `!mounted` that diverges

Ideally, this rule should also be customizable to detect/forbid usages other than `setState`.

- [x] Warns about a potential error (problem)
- [ ] Suggests an alternate way of doing something (suggestion)
- [ ] Other (please specify:)

In methods:

```dart
class FooState extends State {
  fetchData() async {
    final data = await ...;
    setState(() { ... });
  }
}
```

In closures:

```dart
Widget build(context) {
  return FooWidget(
    ..
    onChange: (value) async {
      await fetchData(...);
      setState(() { ... });
    }
  );
}
```

Conditional await (would `mounted` checking be mandatory in this case?)
```dart
void fetchCacheableData() async {
  final data = someCondition ? fromCache : await ...;
  setState(() { ... });
}
```

Yes

Desdaemon:

- [ ] Documentation update
- [ ] Bug fix
- [x] New rule
- [ ] Add a CLI option
- [ ] Other, please explain:

#1117

- [x] Basic implementation
- [x] Rule customization
- [x] Flow analysis
  - [x] `if` checks
  - [x] `while` checks
  - [ ] ~~Ternaries~~
  - [x] And-chains (`mounted && ..`)
  - [x] Or-chains (`!mounted || ..`)

Should this rule try to be as comprehensive as possible and check for all cases, or should the current heuristics for determining mounted-ness be sufficient?


incendial:
@Desdaemon could you please update the changelog?

> Should this rule try to be as comprehensive as possible and check for all cases, or should the current heuristics for determining mounted-ness be sufficient?

The most important that the rule shouldn't produce false-positives. So, if I'm not mistaken, covering `And-chains (mounted && ..)` is a must.

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_async_setstate/avoid_async_setstate_rule.dart
```diff
@@ -0,0 +1,42 @@
...
+import '../../../models/internal_resolved_unit_result.dart';
+import '../../../models/issue.dart';
+import '../../../models/severity.dart';
+import '../../models/flutter_rule.dart';
+import '../../rule_utils.dart';
+
+part 'visitor.dart';
+part 'helpers.dart';
+
+class AvoidAsyncSetStateRule extends FlutterRule {
```
incendial, line 16: Thinking more about the name and looks like it can confuse people like it bans `setState(async () {});`. Maybe you have other ideas how to name it?

incendial (commented):
Great result! Left a few comments. Haven't seen `@pragma('vm:prefer-inline')` in action before 🙂

lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_async_setstate/avoid_async_setstate_rule.dart
Desdaemon: We could follow the precedence here and call it `use-setstate-synchronously`, what do you think?

incendial: Yeah, IMO, this one is better 👍

test/src/analyzers/lint_analyzer/rules/rules_list/use_setstate_synchronously/use_setstate_synchronously_rule_test.dart
```diff
@@ -0,0 +1,47 @@
...
+
+void main() {
+  group('UseSetStateSynchronouslyTest', () {
+    test('initialization', () async {
+      final unit = await RuleTestHelper.resolveFromFile(_examplePath);
+      final issues = UseSetStateSynchronouslyRule().check(unit);
+
+      RuleTestHelper.verifyInitialization(
+        issues: issues,
+        ruleId: 'avoid-async-setstate',
```
incendial, line 17: Please update the name here too

incendial (commented):
Amazing! The only question I have - will `||` create false positives? I'll later run this rule on Flutter codebase (or you can, if you want to) to know for sure, but it'd be better if we avoid as many false positives as possible

Desdaemon:
There's one I haven't found a way to cleanly implement:

```dart
if (!mounted || foo) return;
setState(); // should be mounted here

if (!mounted && bar || foo) return;
setState(); // LINT
```

When the first or-chain diverges, it should also have the effect of guaranteeing the widget is mounted, but currently it does not do so. If there are other false positives I'll prioritize fixing them as well.

Desdaemon:
> I'll later run this rule on the Flutter codebase (or you can, if you want to) to know for sure, but it'd be better if we avoid as many false positives as possible

I ran this on the Flutter package on the stable branch, the only violation it detected was this situation:

https://github.com/flutter/flutter/blob/135454af32477f815a7525073027a3ff9eff1bfd/packages/flutter/lib/src/material/refresh_indicator.dart#L432-L439

I think this is too specific for this rule to handle, but otherwise no false positives yet. I'll run this on other codebases as well to cover as many edge cases as possible.

incendial:
> I think this is too specific for this rule to handle, but otherwise no false positives yet.

Yeah, looks questionable, tbh.

> I'll run this on other codebases as well to cover as many edge cases as possible.

I've noticed a fix for a false positive, if you can also add tests - that'd be great. And after that I'm ready to merge, so tell me when you're ready.

Desdaemon:
@incendial That should be my last batch of changes, should be ready to review and merge.

incendial:
@Desdaemon please add tests for the config ([example](https://github.com/dart-code-checker/dart-code-metrics/blob/master/test/src/analyzers/lint_analyzer/rules/rules_list/avoid_returning_widgets/avoid_returning_widgets_rule_test.dart#L62)) and I'll merge

incendial:
And please check the failing tests

incendial (approved):

incendial:
@Desdaemon thank you for the contribution! 🚀

PR #1123, opened 2022-12-26
fix: correctly handle prefixed enums and static instance fields for prefer-moving-to-variable

incendial:

- [ ] Documentation update
- [X] Bug fix
- [ ] New rule
- [ ] Add a CLI option
- [ ] Other, please explain:


dkrutskikh (approved):

test/src/analyzers/lint_analyzer/rules/rules_list/prefer_moving_to_variable/examples/prefix_example.dart
```diff
@@ -0,0 +1,13 @@
+import 'example.dart' as prefix;
+import 'generics_example.dart';
+
+void main() {
+  AnotherEnum.anotherValue;
+  AnotherEnum.firstValue;
+
+  prefix.SomeValue.firstValue;
```
jacob314, line 8: try referencing prefix.SomeValue.firstValue
repeatedly
I believe that would match the case where I was seeing false positives.

incendial: Yes, that's it! Thank you!

incendial: Hm, actually, if I duplicate the line 5 `AnotherEnum.anotherValue;` the rule will also trigger. So it's not connected to prefixes, maybe you remember another cases?

jacob314: The strange thing is removing the prefixes did seem to fix the issue in DevTools.  That said, I wouldn't want the rule to trigger for duplicated `AnotherEnum.anotherValue` case either.  Specifying the enum value name twice seems more idiomatic than defining an unneeded local variable.

incendial: Agree. And maybe the same should work for the class level static variables. You probably wouldn't want to assign this type of calls to a variable.

jacob314: Agree that I wouldn't want this firing for class to class level static variables either.

incendial: Sorry for misleading, it was a problem exactly with the prefixes 🤦 
Now fixed, thank you!

jacob314 (approved):
lgtm

dkrutskikh (approved):

PR #1166, opened 2023-01-27
feat: add avoid-creating-vector-in-update

incendial:

- [ ] Documentation update
- [ ] Bug fix
- [X] New rule
- [ ] Add a CLI option
- [ ] Other, please explain:


lib/src/analyzers/lint_analyzer/rules/rules_list/avoid_creating_vector_in_update/avoid_creating_vector_in_update_rule.dart
```diff
@@ -0,0 +1,46 @@
...
+import '../../../models/severity.dart';
+import '../../models/flame_rule.dart';
+import '../../rule_utils.dart';
+
+part 'visitor.dart';
+
+class AvoidCreatingVectorInUpdateRule extends FlameRule {
+  static const String ruleId = 'avoid-creating-vector-in-update';
+
+  static const _warningMessage = "Avoid creating Vector2 in 'update' method.";
```
dkrutskikh, line 22: @incendial why we support only Vector2? Please take a look https://pub.dev/documentation/flame/latest/input/Vector-class.html

dkrutskikh (changes requested):

Issue #1178, opened 2023-02-01
[BUG] `use-setstate-synchronously` doesn't recognize `assert` statements


DetachHead:
### Environment and configuration

DCM version: 5.5.1
Dart SDK version:  3.0.0-52.0.dev

<details>
<summary>Configuration</summary>

```yaml
dart_code_metrics:
  extends:
    - package:dart_code_metrics/presets/all.yaml
```
</details>

### What did you do?

```dart
await doThing();
assert(mounted);
setState(() => {}); // error: use-setstate-synchronously
```

### What did you expect to happen?

no error

### What actually happened?

`use-setstate-synchronously` error

### Participation

- [ ] I am willing to submit a pull request for this issue.

incendial:
Oh, considering you have 3.0.0, can you share how is your experience with DCM so far?

DetachHead:
I haven't encountered any problems, although I didn't use it for very long. I was just testing something unrelated and forgot to switch back

incendial:
Still, that's good to know, thank you!

incendial:
@Desdaemon hey, do you want to take a look or should I? If you don't have time, that's totally okay 👍

Desdaemon:
> @Desdaemon hey, do you want to take a look or should I? If you don't have time, that's totally okay 👍 

That'd be great, thanks. I've been having trouble finding free time recently.

incendial:
Available in 5.6.0 🚀

PR #1181, opened 2023-02-03
fix: support assert(mounted) for use-setstate-synchronously

incendial:

#1178

### DIFF
--- a/lib/src/analyzers/lint_analyzer/rules/rules_list/use_setstate_synchronously/visitor.dart
+++ b/lib/src/analyzers/lint_analyzer/rules/rules_list/use_setstate_synchronously/visitor.dart
@@ -44,6 +44,13 @@ class _AsyncSetStateVisitor extends RecursiveAstVisitor<void> {
   }
 
   @override
+  void visitAssertStatement(AssertStatement node) {
+    final newMounted = _extractMountedCheck(node.condition);
+    mounted = newMounted.or(mounted);
+    super.visitAssertStatement(node);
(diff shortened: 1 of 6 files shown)


incendial:
@Desdaemon please take a look, if you have a moment

lib/src/analyzers/lint_analyzer/rules/rules_list/use_setstate_synchronously/visitor.dart
```diff
@@ -65,7 +72,6 @@ class _AsyncSetStateVisitor extends RecursiveAstVisitor<void> {
       return node.visitChildren(this);
     }
 
-    node.condition.visitChildren(this);
```
incendial, line 68: @Desdaemon removing this looks safe or am I missing something?

Desdaemon: This might cause lints to fail for e.g. `if (await ..)`, but I was mostly being cautious when I wrote this.

incendial: Hm, I'll add some tests to check this case

incendial: Actually, it was not checked coz of `.visitChildren`, I've added a test and fixed the problem

Issue #1027, opened 2022-10-10
[New rule] prefer-define-hero-tag


lsaudon:
**Please describe what the rule should do:**

Please describe what the rule should do:
The rule should tell us to give a `heroTag` to the widget that supports it to avoid a `FlutterError`.

**What category of rule is this? (place an "X" next to just one item)**

- [x] Warns about a potential error (problem)
- [ ] Suggests an alternate way of doing something (suggestion)
- [ ] Other (please specify:)

**Provide 2-3 code examples that this rule will warn about (it will be better if you can provide both good and bad examples):**

**❌ Bad:**
```dart
FloatingActionButton(
  onPressed: () => context.read<CounterCubit>().increment(),
  child: const Icon(Icons.add),
), // LINT
```

**✅ Good:**
```dart
FloatingActionButton(
  heroTag: 'increment_heroTag',
  onPressed: () => context.read<CounterCubit>().increment(),
  child: const Icon(Icons.add),
),
```

**Are you willing to submit a pull request to implement this rule?**

No, too complex for me.

incendial:
@lsaudon I thought that this tag is an optional thing and the rule will lead to a lot of false-positives. Or l'm wrong?

lsaudon:
When `heroTag` is not set, then a default `heroTag` is set.
If you have 2 widgets with default `heroTag`, when you go on the page, you have an error.
I don't know about false positives.

incendial:
Available in 5.7.0 🚀

PR #1184, opened 2023-02-04
feat: add prefer-define-hero-tag rule

san-smith:
Issue: #1027 

Added new rule `prefer-define-hero-tag`.

### Good example
```dart
FloatingActionButton(
  heroTag: 'increment_heroTag',
  onPressed: () => context.read<CounterCubit>().increment(),
  child: const Icon(Icons.add),
),
```

### Bad example
```dart
FloatingActionButton(
  onPressed: () => context.read<CounterCubit>().increment(),
  child: const Icon(Icons.add),
), // LINT
```


lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_define_hero_tag/visitor.dart
```diff
@@ -0,0 +1,31 @@
+part of 'prefer_define_hero_tag_rule.dart';
+
+const _floatingActionButtonClassName = 'FloatingActionButton';
```
incendial, line 3: Is the rule really limited only to check for `FloatingActionButton`? There are no other widgets with `heroTag`?

test/src/analyzers/lint_analyzer/rules/rules_list/prefer_define_hero_tag/prefer_define_hero_tag_rule_test.dart
```diff
@@ -0,0 +1,61 @@
...
+    'PreferDefineHeroTagRule',
+    () {
+      test('initialization', () async {
+        final unit = await RuleTestHelper.resolveFromFile(_examplePath);
+        final issues = PreferDefineHeroTagRule().check(unit);
+
+        RuleTestHelper.verifyInitialization(
+          issues: issues,
+          ruleId: 'prefer-define-hero-tag',
+          severity: Severity.performance,
```
incendial, line 20: Why `performance`? According to the rule issue it should have a `warning` severity instead.

test/src/analyzers/lint_analyzer/rules/rules_list/prefer_define_hero_tag/examples/example.dart
```diff
@@ -0,0 +1,93 @@
...
+          onPressed: () {},
+        ),
+      );
+}
+
+class MyWidget9 extends StatelessWidget {
+  const MyWidget9({super.key});
+
+  @override
+  Widget build(BuildContext context) => Scaffold(
```
incendial, line 87: Please add other classes (since we don't have a real flutter dependency), like Widget, StatelessWidget, BuildContext so they are defined in the scope.

lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_define_hero_tag/visitor.dart
```diff
@@ -0,0 +1,31 @@
...
+    super.visitMethodInvocation(node);
+
+    final methodName = node.methodName.name;
+    if (methodName == _floatingActionButtonClassName ||
+        node.beginToken.lexeme == _floatingActionButtonClassName &&
+            (methodName == _constructorExtendedName ||
+                methodName == _constructorLargeName ||
+                methodName == _constructorSmallName)) {
+      if (!node.argumentList.arguments
+          .cast<NamedExpression>()
```
incendial, line 25: Let's get rid of `cast` and check for type in `any`. WDYT?

incendial (commented):
Great job! Left a few comments though.

lib/src/analyzers/lint_analyzer/rules/rules_list/prefer_define_hero_tag/visitor.dart
lsaudon: Hero Widget ?

san-smith: Hero widget has required field `tag` and does not need this rule.
On the other hand, there are two more widgets with field `heroTag`: `CupertinoNavigationBar` and `CupertinoSliverNavigationBar`.

incendial: Maybe we should check all of them with this rule?

san-smith: Yes, I added check for them.

incendial (approved):
Thank you for the contribution!

PR #1190, opened 2023-02-12
fix: correctly handle implicit type parameters for no-equal-arguments

incendial:

Fix implicit type parameters for `no-equal-arguments`.


lib/src/analyzers/lint_analyzer/rules/rules_list/no_equal_arguments/visitor.dart
```diff
@@ -39,14 +39,18 @@ class _Visitor extends RecursiveAstVisitor<void> {
             arg is NamedExpression &&
             argument.expression is! Literal &&
             arg.expression is! Literal) {
-          return argument.expression.toString() == arg.expression.toString();
+          return argument.expression.staticParameterElement?.type ==
```
dkrutskikh, line 43: @incendial I think we need extract compare function like this

```dart
bool compare(Expression a, Expression b) => a.staticParameterElement?.type == b.staticParameterElement?.type && a.toString() == b.toString();
```

dkrutskikh (changes requested):

